PR #16, opened 2026-02-10
Refactor parameters and list modules

ztgrace:
## Summary
- Introduce typed parameter system (`plugin.NewParam[T]`) with validation, defaults, shortcodes, and enum constraints
- Extract CloudControl list logic into reusable `pkg/aws/cloudcontrol` package (`ListAll`, `ListByType`)
- Extract AWS resource type catalogs into `pkg/aws/resourcetypes`
- Simplify `list` and `list-all` modules by delegating to shared packages
- Remove dead code from `pkg/output/types.go` and inline helpers
- Add unit tests for cloudcontrol list, resource types, and list-all module

## Test plan
- [ ] Verify `go build ./...` compiles cleanly
- [ ] Verify `go test ./...` passes
- [ ] Spot-check `list` and `list-all` modules against a live AWS account


pkg/aws/cloudcontrol/list.go
```diff
@@ -0,0 +1,112 @@
...
+	"sync"
+
+	"github.com/aws/aws-sdk-go-v2/service/cloudcontrol"
+	"github.com/praetorian-inc/aurelian/internal/helpers"
+	"github.com/praetorian-inc/aurelian/pkg/output"
+	"github.com/praetorian-inc/aurelian/pkg/ratelimit"
+	"golang.org/x/sync/errgroup"
+)
+
+// ListAll enumerates multiple resource types concurrently with rate limiting.
```
josephwhenry, line 24: Can we make this a struct instead of a 6-parameter function?

pkg/aws/cloudcontrol/list.go
```diff
@@ -0,0 +1,112 @@
...
+// Uses best-effort enumeration: skippable errors (access denied, unsupported types)
+// are logged and skipped. Only context cancellation propagates as an error.
+func ListAll(ctx context.Context, client *cloudcontrol.Client, resourceTypes []string, accountID, region string, concurrency int) (map[string][]output.CloudResource, error) {
```
josephwhenry: We could move `ctx`, `client`, `accountID`, `region`, and `concurrency` to the base type.

Then, we might be able to combine `ListAll` and `List` into one method by turning `resourceTypes` into a variadic parameter.

pkg/plugin/module.go
```diff
@@ -39,17 +42,276 @@ type Parameter struct {
...
+	return ps.params
+}
+
+// Len returns the number of parameters.
+func (ps *Parameters) Len() int {
+	return len(ps.params)
+}
+
+// String returns the effective string value for the named parameter.
+func (ps *Parameters) String(name string) string {
```
josephwhenry: These all look pretty similar. Can we consolidate these behind a helper that uses generics?

pkg/aws/cloudcontrol/list.go
ztgrace: I merged the opts into a struct but left ctx and client separate.

pkg/plugin/module.go
ztgrace: I cleaned this up.

PR #25, opened 2026-02-13
Cloud Control Tweaks

josephwhenry:


pkg/aws/cloudcontrol/list_integration_test.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+}
+
+func verifyRegionPresent(t *testing.T, results map[string][]output.CloudResource, region string) {
+	t.Helper()
+
+	regionFound := false
+	for key, result := range results {
+		if !strings.HasPrefix(key, region) {
+		}
```
ztgrace, line 63:
```suggestion
		if !strings.HasPrefix(key, region) {
		    continue
		}
```

pkg/ratelimit/paginator.go
```diff
@@ -0,0 +1,67 @@
...
+	return p.ShouldRetry(err)
+}
+
+func (p *Paginator) backoff(retryAttempt int) {
+	if p.Backoff != nil {
+		p.Backoff(retryAttempt)
+		return
+	}
+	// default backoff
+	time.Sleep(5 * time.Duration(1<<retryAttempt))
```
ztgrace:
```suggestion
	time.Sleep(5 * time.Second * time.Duration(1<<retryAttempt))
```

time.Duration's base unit is nanoseconds

pkg/aws/cloudcontrol/list_new.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	accountID, err := cc.getAccountID(region)
+	if err != nil {
+		return nil, err
+	}
+
+	var all []output.CloudResource
+	for _, resourceType := range resourceTypes {
+		resources, err := cc.listByType(client, accountID, region, resourceType)
```
ztgrace: sequential is ok for now, but we should think about concurrency here.

pkg/aws/cloudcontrol/list_new.go
```diff
@@ -0,0 +1,212 @@
...
+	awshelpers "github.com/praetorian-inc/aurelian/internal/helpers/aws"
+	"github.com/praetorian-inc/aurelian/pkg/output"
+	"github.com/praetorian-inc/aurelian/pkg/plugin"
+	"github.com/praetorian-inc/aurelian/pkg/ratelimit"
+)
+
+type CloudControlLister struct {
+	Concurrency      int
+	Profile          string
+	ProfileDir       string
```
ztgrace:
```suggestion
	AWSCommonRecon
```

Any reason we can't use the common recon params? That struct needs to be cleaned up since we removed some of the functionality, but otherwise param details won't get populated to the cli

PR #34, opened 2026-02-16
bind cleanup

josephwhenry:


pkg/testutils/mock_module.go
```diff
@@ -30,7 +30,7 @@ func (m *MockModule) OpsecLevel() string {
 }
 func (m *MockModule) Authors() []string              { return m.AuthorsValue }
 func (m *MockModule) References() []string           { return m.ReferencesValue }
-func (m *MockModule) Parameters() []plugin.Parameter { return m.ParametersValue }
+func (m *MockModule) Parameters() any { return m.ParametersValue }
```
ztgrace, line 33: I'm not a fan of this, we lose type safety. I'd much rather have the bind boiler plate, I don't wany any of the module typing left up to runtime validation

PR #38, opened 2026-02-18
adding logging/cli handling

ztgrace:


pkg/modules/aws/recon/list_all_resources.go
```diff
@@ -59,8 +59,18 @@ func (m *AWSListAllResourcesModule) Run(cfg plugin.Config) ([]plugin.Result, err
...
+	resourceTypes := selectResourceTypes(c.ScanType)
+
 	lister := cclist.NewCloudControlLister(c.AWSCommonRecon)
-	results, err := lister.List(resolvedRegions, selectResourceTypes(c.ScanType))
+
+	if cfg.Log != nil {
+		cfg.Log.Info("scanning %d regions for %d resource types", len(resolvedRegions), len(resourceTypes))
+		p := cfg.Log.Progress("scanning resources", len(resolvedRegions)*len(resourceTypes))
+		lister.OnProgress = func() { p.Increment() }
+		defer p.Done()
```
josephwhenry: Should we just pass an optional logger to the CloudControlLister instead of using an optional callback here? I think that would be easier to follow.

pkg/aws/cloudcontrol/list.go
```diff
@@ -52,6 +53,9 @@ func (cc *CloudControlLister) List(identifier string, out *pipeline.P[output.AWS
 }
 
 func (cc *CloudControlLister) ListByARN(resourceARN string, out *pipeline.P[output.AWSResource]) error {
+	if cc.OnProgress != nil {
+		defer func() { cc.OnProgress(resourceARN) }()
+	}
```
josephwhenry: I bet we can build this into the Pipeline pattern somehow, so each component just exposes some PeriodicLog or OnProgress method or something, and then pipeline.Pipe calls it after processing each item.

pkg/modules/aws/recon/account_auth_details.go
```diff
@@ -53,12 +53,21 @@ func (m *AWSAccountAuthDetailsModule) Parameters() any {
 }
 
 func (m *AWSAccountAuthDetailsModule) Run(cfg plugin.Config, out *pipeline.P[model.AurelianModel]) error {
+	if cfg.Log != nil {
+		cfg.Log.Info("fetching account authorization details")
+	}
```
josephwhenry: We should either assume cfg always has a `Log`, or else tweak how we pass `Log` into `Cfg` so that we can run the `Info`, etc. methods without risking a `nil` crash.

For example:
```
func (cfg *Config) Info(msg string) {
    if cfg.Log != nil {
        cfg.Log.Info(msg)
    }
}
```

That will let us just call `cfg.Info` or whatever without needing to always check if `Log` is nil.

pkg/modules/aws/recon/find_secrets.go
```diff
@@ -86,12 +86,31 @@ func (m *AWSFindSecretsModule) Run(cfg plugin.Config, out *pipeline.P[model.Aure
...
+	if cfg.Log != nil {
+		progress = cfg.Log.Progress("listing", total)
+	}
+
 	lister := cclist.NewCloudControlLister(c.AWSCommonRecon)
+	lister.OnProgress = func(detail string) {
+		if progress != nil {
+			progress.IncrementWithDetail(detail)
+		}
+	}
```
josephwhenry: I would be very careful to avoid cluttering the modules with various logging features. That's going to make them very difficult for a human or agent to understand. 

We should push as much of this to the components (e.g., CloudControlLister) or the `pipeline.Pipe` function as possible. Module methods is another decent option.

josephwhenry (changes requested):

pkg/aws/cloudcontrol/list.go
ztgrace: implemented

pkg/modules/aws/recon/find_secrets.go
ztgrace: all is handled in the pipeline now.

PR #39, opened 2026-02-18
Add AWS Attack Graph to Guard

josephwhenry:


pkg/modules/aws/analyze/analyze_iam_permissions.go
```diff
@@ -108,12 +109,37 @@ func (m *AnalyzeIAMPermissionsModule) Run(cfg plugin.Config) ([]plugin.Result, e
 
 	results := analyzer.FullResults(summary)
 
+	// Convert GAAD entities to AWSIAMResource
+	var entities []output.AWSIAMResource
```
ztgrace: Let's create a helper for this since it's duplicated in graph.go

pkg/output/aws_iam_resource.go
```diff
@@ -0,0 +1,37 @@
...
+	IAMTags                 any      `json:"iam_tags,omitempty"`
+
+	// OriginalData holds the original GAAD struct (UserDL, RoleDL, etc.)
+	// for lossless conversion back to graph nodes. Not serialized to JSON.
+	OriginalData any `json:"-"`
+}
+
+// IsIAMResource returns true if this resource has IAM-specific fields populated.
+func (r *AWSIAMResource) IsIAMResource() bool {
+	return r.InlinePolicies != nil ||
```
ztgrace: should we check for permissionsboundary and iamtags too?

pkg/output/aws_iam_resource.go
```diff
@@ -0,0 +1,37 @@
+package output
+
+// AWSIAMResource wraps CloudResource with typed IAM fields.
+// For non-IAM resources, all IAM-specific fields are nil/empty.
+// This is an output-boundary type — internal analyzer types remain unchanged.
+type AWSIAMResource struct {
+	CloudResource
+
+	// IAM-specific typed fields (nil/empty for non-IAM resources)
+	InlinePolicies          any      `json:"inline_policies,omitempty"`
```
ztgrace: can we set this to a concrete type? We may need to reorganize to prevent import cycles but I think that's better than using any

ztgrace (changes requested):

PR #42, opened 2026-02-19
Add AWS Attack Graph to Guard (Refactor)

josephwhenry:


pkg/aws/iam/gaad/analyzer.go
```diff
@@ -0,0 +1,129 @@
...
+
+// NewGaadAnalyzer creates a new GaadAnalyzer.
+func NewGaadAnalyzer() *GaadAnalyzer {
+	return &GaadAnalyzer{}
+}
+
+// Analyze evaluates all principal permissions against resources and org policies,
+// detecting privilege escalation paths. Returns the set of allowed permission
+// edges (principal → action → resource).
+func (ga *GaadAnalyzer) Analyze(
```
josephwhenry, line 27: @ztgrace Can you walk through this code (at least at a high level) to ensure I didn't radically screw up the process? My mental model of the old GAAD analyzer was:
1. Concurrency: "processor" and "evaluator" goroutines. The former generate events (based on real resources) for the latter to process.
2. Run goroutines for the following "processors":
    - IAM users
    - IAM roles
    - Resource policies (specifically, looking for service principals that can act on resources)
    - AssumeRole policies (special case of IAM roles)
3. Pass everything to the evaluator (didn't dig too much into the evaluator, that code is mostly untouched)
4. For each result flagged by the evaluator, create some object that represents the "relationship"

pkg/aws/iam/service_action_resource_map.go
```diff
@@ -97,7 +97,7 @@ var serviceResourceMaps = map[string]ServiceResourceMap{
 			"oidc-provider":      regexp.MustCompile(`^arn:aws:iam::\d{12}:oidc-provider/.*`),
 			"saml-provider":      regexp.MustCompile(`^arn:aws:iam::\d{12}:saml-provider/.*`),
 			"server-certificate": regexp.MustCompile(`^arn:aws:iam::\d{12}:server-certificate/.*`),
-			"service":            regexp.MustCompile(`^iam.amazonaws.com$`),
+			"service":            regexp.MustCompile(`^arn:aws:iam:\*:\*:\*$`),
```
josephwhenry, line 100: @ztgrace Can you also sanity check this? I found it simpler to just always use dummy ARNs for services, instead of having a bunch of edge cases for AWS services.

PR #57, opened 2026-02-26
Add EC2 IMDSv2 enforcement check via enricher and YAML rule

bernard-yip:
## Summary
- Adds an enricher (`ec2_imds.go`) that calls `ec2:DescribeInstances` to fetch `MetadataOptions` for each EC2 instance, since CloudControl's `ListResources` doesn't return this data
- Adds a YAML rule (`ec2-imdsv1-enabled.yaml`) that flags running instances with IMDS enabled but IMDSv1 not disabled (HttpTokens != required)
- Includes unit tests (6 cases) and integration test with Terraform infra (3 EC2 instances: IMDSv1 allowed, IMDSv2 enforced, IMDS disabled)

## Dependencies
- **Integration tests require a fix to `test/testutil/fixture.go`** — the hardcoded state bucket name `aurelian-integration-tests` is owned by another AWS account (403). A separate PR will add `resolveStateBucket()` using STS `GetCallerIdentity` to dynamically construct the bucket name.

## Linear
Closes LAB-1198

## Test plan
- [x] `go build` compiles cleanly
- [x] Unit tests pass (`go test ./pkg/modules/aws/enrichers/...`)
- [x] Integration test passes: enrichment adds IMDS properties to all 3 instances
- [x] Integration test passes: YAML rule flags only IMDSv1-allowed instance, zero false positives

### DIFF
--- a/pkg/modules/aws/enrichers/ec2_imds.go
+++ b/pkg/modules/aws/enrichers/ec2_imds.go
@@ -1,0 +1,67 @@
+package enrichers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	smithy "github.com/aws/smithy-go"
+	"github.com/praetorian-inc/aurelian/pkg/output"
+	"github.com/praetorian-inc/aurelian/pkg/plugin"
+)
+
+func init() {
+	plugin.RegisterEnricher("AWS::EC2::Instance", fetchIMDSMetadataWrapper)
+}
+
+// EC2DescribeInstancesAPI is the minimal interface needed for testing.
+type EC2DescribeInstancesAPI interface {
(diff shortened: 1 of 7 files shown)


bernard-yip:
**Note:** Integration tests for this PR require a fix to `test/testutil/fixture.go` — the hardcoded S3 state bucket name `aurelian-integration-tests` returns 403 (owned by another account). A separate PR will add `resolveStateBucket()` using STS `GetCallerIdentity` to dynamically construct the bucket name as `aurelian-integration-tests-<account-id>`.

Without that fix, `go test -tags=integration` will fail at the `verifyStateBucket` step.

Linear: https://linear.app/praetorianlabs/issue/LAB-780/non-mvp-aws-attack-capabilities-aurelian

ztgrace (approved):

PR #60, opened 2026-03-02
initial find secrets implementation

josephwhenry:


pkg/secrets/scanner.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Scan is a pipeline-compatible method that scans a ScanInput for secrets
+// and sends SecretScanResult values to the output pipeline.
+func (s *SecretScanner) Scan(input output.ScanInput, out *pipeline.P[SecretScanResult]) error {
+	blobID := types.ComputeBlobID(input.Content)
+	provenance := types.FileProvenance{FilePath: input.Label}
+
+	matches, err := s.ps.scanContent(input.Content, blobID, provenance)
+	if err != nil {
+		return nil // skip content that fails to scan
```
ztgrace: we should at least log this error

```suggestion
        slog.Warn("failed to scan content for secrets", "resource", input.ResourceID, "label", input.Label, "error", err)
		return nil
```

PR #63, opened 2026-03-02
feat: add public AMI detection module

CCushing-Praetorian:
## Summary

- Adds `public-ami` recon module that detects account-owned AMIs with public launch permissions (`Group: all`)
- Scans across regions concurrently using `ratelimit.NewAWSRegionLimiter` + errgroup
- Classifies risks as **High** (public AMI in use by running instances) or **Medium** (public but stale)
- Extracts associated EBS snapshot IDs and tracks instance usage
- Adds `model.BaseAurelianModel` embed to `output.Risk` for pipeline compatibility (same change as CDK branch)

## Files

| File | Purpose |
|------|---------|
| `pkg/aws/ec2/ami/types.go` | `AMIInfo`, `ScanOptions`, `ScanResult` |
| `pkg/aws/ec2/ami/scan.go` | Multi-region scan orchestration |
| `pkg/aws/ec2/ami/detection.go` | EC2 client interface, launch permission checks, instance usage |
| `pkg/aws/ec2/ami/risks.go` | Risk generators (TH in-use, TM stale) |
| `pkg/aws/ec2/ami/scan_test.go` | 10 unit tests with mocked EC2 client |
| `pkg/modules/aws/recon/public_ami.go` | Thin recon module (~68 lines) |
| `pkg/output/types.go` | `BaseAurelianModel` embed on `Risk` |

## Test plan

- [x] 10 unit tests pass (public in-use, stale, private, no AMIs, multiple AMIs, error handling, snapshot extraction, risk field validation)
- [x] Integration tested against sandbox (account `914697327504`):
  - Created public AMI + running instance → detected as TH `public-ami-in-use`
  - Terminated instance, re-scanned → detected as TM `public-ami-stale`
  - Private AMI correctly excluded (true negative)
- [x] All existing tests pass (`pkg/modules/aws/recon/`, `pkg/output/`)
- [x] `go build` clean (pre-existing titus issue only)

Resolves: LAB-698

### DIFF
--- a/pkg/aws/enumeration/ec2_image_enumerator.go
+++ b/pkg/aws/enumeration/ec2_image_enumerator.go
@@ -1,0 +1,210 @@
+package enumeration
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"slices"
+	"strings"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awsarn "github.com/aws/aws-sdk-go-v2/aws/arn"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/praetorian-inc/aurelian/pkg/output"
+	"github.com/praetorian-inc/aurelian/pkg/pipeline"
+	"github.com/praetorian-inc/aurelian/pkg/plugin"
+	"github.com/praetorian-inc/aurelian/pkg/ratelimit"
+)
+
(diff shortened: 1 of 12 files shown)


ztgrace:
I got this cleaned up with the latest version of main. But I think we need to merge this into the standard public resource module. But the challenge is that AMIs aren't a supported cloudcontrol resource type, so we'll need to come up with a solution for that first. I'll come back to this once we solve that.

ztgrace (approved):

PR #64, opened 2026-03-03
Public Resource Module (Part 2)

josephwhenry:


pkg/plugin/aws_params.go
```diff
@@ -50,26 +50,36 @@ func (c *AWSCommonRecon) PostBind(_ Config, _ Module) error {
...
+		return fmt.Errorf("loading org policies: %w", err)
+	}
+
+	c.OrgPolicies = op
 	return nil
 }
+
+type ResourceARNParam struct {
+	ResourceARN []string `param:"resource-arn"   desc:"AWS target resource ARN" shortcode:"a"`
+}
```
ztgrace: I like the mulit-arn support but think this needs to get moved to he AWSCommonRecon struct.

PR #71, opened 2026-03-04
Azure List All (Part 3)

josephwhenry:


pkg/modules/azure/recon/list_all_resources.go
```diff
@@ -44,34 +55,53 @@ func (m *AzureListAllResourcesModule) References() []string {
...
+		return nil
 	}
 
-	lister, err := resourcegraph.NewResourceGraphLister(cred, nil)
+	idStream := pipeline.From(subscriptionIDs...)
+	resolvedSubs := pipeline.New[azuretypes.SubscriptionInfo]()
+	pipeline.Pipe(idStream, resolver.Resolve, resolvedSubs)
+
+	lister := resourcegraph.NewResourceGraphLister(m.AzureCredential, nil)
+	pipeline.Pipe(resolvedSubs, lister.ListAll, resources)
```
josephwhenry, line 85: I think this is a stronger pipeline pattern than what was there before. Each components exposes a pipeline-compatible method, and we can easily feed each subscription into the lister. 

Not a huge deal for this module, but it will set a good pattern for later modules.

pkg/modules/azure/recon/list_all_resources_test.go
```diff
@@ -1,35 +0,0 @@
-package recon
```
josephwhenry, line 1: I removed this test, as it doesn't provide anything useful that out integration test won't check. And most of the stuff here is brittle (e.g., if we tweak the module name or a parameter, now this test will fail).

pkg/plugin/azure_params.go
```diff
@@ -1,6 +1,24 @@
...
 	SubscriptionID []string `param:"subscription-id" desc:"Azure subscription ID(s) or 'all' to enumerate all accessible subscriptions" default:"all" shortcode:"s"`
+	AzureCredential azcore.TokenCredential `param:"-"`
+}
+
+func (c *AzureCommonRecon) PostBind(_ Config, _ Module) error {
+	cred, err := azureauth.NewAzureCredential()
+	if err != nil {
+		return fmt.Errorf("azure authentication failed: %w", err)
+	}
+	c.AzureCredential = cred
```
josephwhenry, line 22: This seems like boilerplate that we will always want to run before an Azure module, so I moved it to the parameter setup.

pkg/azure/resourcegraph/lister_integration_test.go
```diff
@@ -0,0 +1,66 @@
+//go:build integration
```
josephwhenry, line 1: Replaced the unit test with an integration test. This component sits close enough to the SDK/API, that I don't think we really need to have both.

PR #96, opened 2026-03-12
fix: rewrite azure find_secrets_test.go to match AWS test pattern

### LINKED ISSUE #95: fix: find_secrets_test.go references removed methods, breaks package tests
## Bug

`pkg/modules/azure/recon/find_secrets_test.go` references two methods that no longer exist on `AzureFindSecretsModule`:

- `formatSecretRiskName` 
- `riskSeverityFromMatch`

This causes a build failure for the entire `pkg/modules/azure/recon` package test suite:

```
pkg/modules/azure/recon/find_secrets_test.go:30:13: m.formatSecretRiskName undefined
pkg/modules/azure/recon/find_secrets_test.go:43:45: m.riskSeverityFromMatch undefined
```

## Root Cause

PR #86 added these tests in commit `b4a4c26`, but a subsequent refactor (PR #91) removed/inlined the methods from `find_secrets.go` without updating the test file.

## Impact

All unit tests in `pkg/modules/azure/recon/` fail to compile. Integration tests (behind `//go:build integration`) are unaffected since they compile separately.

## Fix

Either restore the methods as exported helpers, or delete `find_secrets_test.go` if the logic was inlined and is now covered by integration tests.

Elgin-Lee:
## Summary
- `find_secrets_test.go` called `m.formatSecretRiskName()` and `m.riskSeverityFromMatch()` which were removed from `AzureFindSecretsModule` when logic was consolidated into `pkg/secrets/risk.go`
- Rewrites the test to mirror the comprehensive AWS `find_secrets_test.go` pattern instead of just updating the function calls:
  - Module metadata, supported resource types, and parameter validation
  - `ExtractRuleShortName` and risk name formatting with realistic Titus rule IDs
  - `RiskSeverityFromMatch` for all validation states
  - `buildProofData` structural validation (with and without validation result)
  - Proof JSON round-trip serialization
  - `riskFromScanResult` end-to-end with pipeline (ImpactedResourceID with/without FindingID, DeduplicationID, Name, Severity)

## Test plan
- [x] `go test ./pkg/modules/azure/recon/ -v -count=1` — all 17 tests pass (11 new + 6 pre-existing)

Fixes #95

### DIFF
--- a/pkg/modules/azure/recon/find_secrets_test.go
+++ b/pkg/modules/azure/recon/find_secrets_test.go
@@ -1,60 +1,300 @@
 package recon
 
 import (
+	"encoding/json"
+	"fmt"
 	"testing"
 
+	"github.com/praetorian-inc/aurelian/pkg/model"
 	"github.com/praetorian-inc/aurelian/pkg/output"
+	"github.com/praetorian-inc/aurelian/pkg/pipeline"
+	"github.com/praetorian-inc/aurelian/pkg/plugin"
+	"github.com/praetorian-inc/aurelian/pkg/secrets"
 	"github.com/praetorian-inc/titus/pkg/types"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
-func TestFormatSecretRiskName(t *testing.T) {
+func TestFindSecretsModuleMetadata(t *testing.T) {
 	m := &AzureFindSecretsModule{}
+
+	assert.Equal(t, "find-secrets", m.ID())
+	assert.Equal(t, "Azure Find Secrets", m.Name())
+	assert.Equal(t, plugin.PlatformAzure, m.Platform())
+	assert.Equal(t, plugin.CategoryRecon, m.Category())
+	assert.Equal(t, "moderate", m.OpsecLevel())
+
+	authors := m.Authors()
+	require.Len(t, authors, 1)
+	assert.Equal(t, "Praetorian", authors[0])
+
+	assert.NotEmpty(t, m.Description())
+	assert.NotEmpty(t, m.References())
+}
+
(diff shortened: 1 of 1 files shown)


Elgin-Lee:
@ztgrace should we consider not duplicating these tests and centralizing it?

Option 1: Keep the duplicated tests (current state, matches AWS/GCP pattern)

ExtractRuleShortName, RiskNameFormatting, and RiskSeverityFromMatch are tested identically in all three platform test files. Each platform file is self-contained — you can read one file and understand the full contract. If someone changes the secrets package behavior, all three platform tests break, making the blast radius immediately visible. Downside: three copies of the same assertions testing the same exported functions.

Option 2: Deduplicate into pkg/secrets/risk_test.go

Move ExtractRuleShortName, RiskNameFormatting, and RiskSeverityFromMatch tests into the secrets package where the functions live. Each platform test file only tests platform-specific code (buildProofData, riskFromScanResult, module metadata). Cleaner separation of concerns — tests live next to the code they test. Downside: breaks the self-contained pattern AWS established, and you'd need to touch all three platform test files.

Option 3: Shared test helpers imported by each platform

Define the test cases and runner functions once in pkg/secrets/testing_helpers.go, then each platform test file calls them with a one-liner. Test logic is maintained in one place — add a case, all three platforms pick it up. Each platform still runs the tests in its own suite, so go test ./pkg/modules/azure/... breaks if the contract changes, preserving blast radius visibility. The platform-specific parts (like the "azure-secret-" prefix in risk name formatting) get passed as parameters. Downside: it's a new pattern that doesn't exist in the codebase yet, so it requires updating all three platform test files (AWS, GCP, Azure) to be consistent, and it adds an exported test helper API to a non-test file that's only useful for tests.

I assume we are going for 1 to show what broke at the module-level and keep it simple, but just wanted to check

ztgrace (approved):

PR #112, opened 2026-03-16
feat(azure): add subdomain takeover recon module

ztgrace:
## Summary

- Add Azure subdomain takeover detection module that scans DNS zones across subscriptions for dangling records
- Implements registry-based checker pattern (matching AWS module architecture) with 6 checkers:
  - **CNAME**: App Service, Blob Storage, Traffic Manager, CDN Classic — uses Azure `CheckNameAvailability` APIs to confirm names are claimable
  - **A/AAAA**: Orphaned public IP detection via cross-subscription IP enumeration
  - **NS**: Dangling delegation detection via live DNS queries to Azure DNS nameservers
- Includes Terraform fixtures and integration test covering all 6 positive cases + 2 negative cases + structural integrity assertions

## Architecture

Uses the AWS registry pattern (`mustRegister` + `init()`) so new service checkers can be added as single files without touching existing code. Pipeline chain: `subscriptionIDs → enumerator → records → checker → out`.

## Test plan

- [ ] `go vet ./pkg/azure/dnstakeover/... ./pkg/modules/azure/recon/...` passes
- [ ] `go build ./...` succeeds
- [ ] `go test ./pkg/azure/dnstakeover/... ./pkg/modules/azure/recon/...` passes (unit)
- [ ] Integration test passes against provisioned Azure DNS zone with dangling records


Elgin-Lee:
## E2E Test Run Results: 15/16 PASS, 1 FAIL

Ran the full integration test suite against live Azure (`972c9e6c` subscription). All tests pass except **`detects_dangling_Storage_CNAME`**.

### Root Cause: Invalid storage account name in fixture

The storage CNAME fixture uses `${local.prefix}nonexiststor.blob.core.windows.net`, which produces a name like `aursdt-l3ozdcnonexiststor`. This contains a hyphen and is 25 chars — both invalid for Azure storage accounts (must be 3-24 chars, lowercase alphanumeric only):

```
$ az storage account check-name --name "aursdt-l3ozdcnonexiststor"
{
  "nameAvailable": false,
  "reason": "AccountNameInvalid",
  "message": "...must be between 3 and 24 characters in length and use numbers and lower-case letters only."
}
```

The `CheckNameAvailability` API returns `nameAvailable: false` for invalid names, so the checker correctly skips it — it just never gets a chance to detect the takeover.

### Suggested Fix

Add a sanitized prefix (like the `public-resources` fixture does) and use it for the storage CNAME:

```hcl
locals {
  prefix     = "${var.prefix}-${random_string.suffix.result}"
  prefix_san = "${var.prefix}${random_string.suffix.result}"   # <-- add this
}

resource "azurerm_dns_cname_record" "storage_dangling" {
  ...
  record = "${local.prefix_san}noexist.blob.core.windows.net"  # <-- use prefix_san
}
```

### Full Results

| Test | Result |
|---|---|
| detects dangling App Service CNAME | ✅ |
| detects dangling Storage CNAME | ❌ (fixture name invalid) |
| detects dangling Traffic Manager CNAME | ✅ |
| detects dangling CDN CNAME | ✅ |
| detects orphaned IP A record | ✅ |
| detects dangling NS delegation | ✅ |
| safe CNAME does not trigger | ✅ |
| safe A record does not trigger | ✅ |
| 8 structural integrity tests | ✅ all pass |

Everything else looks good — clean implementation that follows existing codebase patterns.

Elgin-Lee (approved):
1 comment then LGTM

PR #115, opened 2026-03-17
fix: eliminate Cartesian product bug in 26 privesc enrichment queries

praetorian-peter-mueller:
The CAN_PRIVESC enrichment queries (methods 01-13, 20-29, 33, 35, 39)
contained a bug where each query dropped its matched target node via
`WITH attacker` and then created CAN_PRIVESC edges to ALL other
principals in the graph — a Cartesian product producing O(n²) false
edges instead of connecting the attacker to its actual permission target.

Fix: replace the `WITH attacker / MATCH (victim:Principal)` pattern
with a direct `MERGE (attacker)-[:CAN_PRIVESC]->(target)` using
`coalesce(target.Arn, target.arn)` to handle both property naming
conventions. Method 39 (lambda UpdateCode+Invoke) preserves target
through its dual-permission check via `WITH attacker, target`.

Tested against a real AWS account (11 users, 52 roles, 37 resources):
- CAN_PRIVESC edges: 1,570 → 744 (52.6% reduction)
- Privesc paths (1-3 hops): 6,030 → 18 (99.7% reduction)
- Same 6 attackers reach same 5 admin targets (no real paths lost)
- Enrichment time: 3.02s → 0.54s (82% faster)

Also adds cmd/graphload/ (dev tool for Neo4j measurement) and tests
for the query pattern correctness.


praetorian-peter-mueller:
@ztgrace @josephwhenry Honestly not sure if the `cmd/graphload/main.go` stuff really belongs in here, it was more of a hack for testing purposes.

PR #120, opened 2026-03-18
Aurelian AWS Privilege Escalation Detection

josephwhenry:


pkg/graph/neo4j/neo4j.go
```diff
@@ -0,0 +1,117 @@
...
+	// fragment given the node's alias. If a Kind is missing, compilation fails.
+	NodePredicates map[string]func(alias string) string
+}
+
+// DefaultNeo4jCompiler returns a compiler with Chariot's standard node-kind mappings.
+func DefaultNeo4jCompiler() *Neo4jCompiler {
+	return &Neo4jCompiler{
+		NodePredicates: map[string]func(alias string) string{
+			"Principal": func(a string) string {
+				return a + "._resourceType IN ['AWS::IAM::User', 'AWS::IAM::Role', 'AWS::IAM::Group']"
```
ztgrace, line 26: the original code always applied a Label of `Principal` to any principal. We should leverage that assumption here rather than these specific labels.

pkg/modules/aws/analyze/detect_privescs.go
```diff
@@ -0,0 +1,168 @@
...
+		return err
+	}
+	defer queryer.Close()
+
+	ctx := context.Background()
+	for _, method := range allMethods() {
+		if err := runMethod(ctx, queryer, method, out); err != nil {
+			slog.Warn("privesc method failed", "id", method.ID(), "error", err)
+		}
+	}
```
ztgrace:
```suggestion
	for _, method := range allMethods() {
		if err := runMethod(ctx, queryer, method, out); err != nil {
			slog.Warn("privesc method failed", "id", method.ID(), "error", err)
		}
	}
	if failures == len(allMethods()) {                     
      return fmt.Errorf("all %d privesc methods failed", failures)                                                                                                                        
    }
```
we should error if all method failed

PR #123, opened 2026-03-19
fix(azure/find-secrets): fix webapp-hostkeys 404, Cosmos cross-partition query, add --max-cosmos-doc-scan

Elgin-Lee:
## Summary

- **webapp-hostkeys 404**: `extractWebAppHostKeys` was calling `ListHostKeys` on all `microsoft.web/sites` resources. Regular Web Apps return 404 because only Function Apps have host keys. Fixed by adding `kind` to ARG query projections and `hydrateFromARG`, then guarding the extractor to skip non-Function Apps when `kind` is known.
- **Cosmos cross-partition 400**: `SELECT TOP 50 * FROM c` is rejected by the Cosmos DB gateway for cross-partition queries — it cannot merge ranked results across physical partitions without client-side execution (not supported by the Go SDK). Fixed by dropping `TOP 50` — the gateway handles fan-out pagination natively without it. `EnableCrossPartitionQuery` removed since the SDK defaults it to `true`.
- **`--max-cosmos-doc-scan`**: Adds an opt-in cap on total documents scanned per Cosmos container. Default is 0 (unlimited — scan all documents). Use when scan time on large containers is a concern.

## Validation

Empirically validated against a deployed test fixture (22 resource types) with a minimum-privilege service principal (Reader + custom role). All 33 expected findings detected, zero 403/400 errors, zero spurious warnings.

## Test plan

- [ ] Run `go test ./pkg/azure/extraction/...` locally
- [ ] Run integration test against fixture with minimum-privilege SP
- [ ] Confirm no `webapp-hostkeys` warnings on non-Function App `microsoft.web/sites` resources
- [ ] Confirm Cosmos config-doc extraction succeeds for partitioned containers

### DIFF
--- a/docs/aurelian_azure_recon_find-secrets.md
+++ b/docs/aurelian_azure_recon_find-secrets.md
@@ -14,6 +14,7 @@ aurelian azure recon find-secrets [flags]
       --disabled-titus-rules strings   Rule IDs to exclude from scanning
   -h, --help                           help for find-secrets
       --ignore-file string             Path to gitignore-style file for skipping paths; empty uses titus defaults
+      --max-cosmos-doc-scan int        Max total Cosmos documents to scan per container (0 = unlimited)
       --max-cosmos-doc-size int        Max individual Cosmos document size in bytes (default 1048576)
       --output-dir string              Base output directory (default "aurelian-output")
   -i, --resource-id strings            Azure resource ID(s) to scan directly, skipping enumeration
--- a/docs/aurelian_azure_recon_find-secrets.md
+++ b/docs/aurelian_azure_recon_find-secrets.md
@@ -14,7 +14,7 @@ aurelian azure recon find-secrets [flags]
       --disabled-titus-rules strings   Rule IDs to exclude from scanning
   -h, --help                           help for find-secrets
       --ignore-file string             Path to gitignore-style file for skipping paths; empty uses titus defaults
-      --max-cosmos-doc-scan int        Max total Cosmos documents to scan per container (0 = unlimited)
+      --max-cosmos-doc-scan int        Max total Cosmos documents to scan per container (default 50)
       --max-cosmos-doc-size int        Max individual Cosmos document size in bytes (default 1048576)
       --output-dir string              Base output directory (default "aurelian-output")
   -i, --resource-id strings            Azure resource ID(s) to scan directly, skipping enumeration
--- a/pkg/azure/extraction/extract_cosmos.go
+++ b/pkg/azure/extraction/extract_cosmos.go
@@ -260,27 +260,36 @@ func extractCosmosConfigDocs(ctx extractContext, r output.AzureResource, out *pi
 }
 
 func queryConfigDocs(ctx extractContext, r output.AzureResource, containerClient *azcosmos.ContainerClient, dbName, containerName string, out *pipeline.P[output.ScanInput]) {
-	query := "SELECT TOP 50 * FROM c"
-	// Use a cross-partition query by providing an empty partition key.
-	crossPartition := true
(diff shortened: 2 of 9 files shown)


pkg/azure/resourcegraph/lister.go
```diff
@@ -65,7 +65,7 @@ func buildFilteredQuery(resourceTypes []string) string {
 	for i, rt := range resourceTypes {
 		quoted[i] = "'" + strings.ToLower(rt) + "'"
 	}
-	return "Resources | where type in~ (" + strings.Join(quoted, ",") + ") | project id, name, type, location, resourceGroup, tags, properties"
+	return "Resources | where type in~ (" + strings.Join(quoted, ",") + ") | project id, name, type, kind, location, resourceGroup, tags, properties"
```
bernard-yip, line 68: `lister_test.go` might break on merge

`TestBuildFilteredQuery` and `TestBuildFilteredQuery_Single` assert exact query strings that no longer match after adding kind to the projection. The PR's test plan says to run `go test ./pkg/azure/extraction/...` but does not mention `./pkg/azure/resourcegraph/...`, so this was likely missed.

pkg/azure/resourcegraph/lister_test.go
Fix: Update expected strings to include kind in the projection.

docs/aurelian_azure_recon_find-secrets.md
```diff
@@ -14,6 +14,7 @@ aurelian azure recon find-secrets [flags]
       --disabled-titus-rules strings   Rule IDs to exclude from scanning
   -h, --help                           help for find-secrets
       --ignore-file string             Path to gitignore-style file for skipping paths; empty uses titus defaults
+      --max-cosmos-doc-scan int        Max total Cosmos documents to scan per container (0 = unlimited)
```
bernard-yip: default to unlimited for number of docs to scan is scary for accounts with large number of docs. maybe default to first 1000?

pkg/azure/extraction/extract_cosmos.go
```diff
@@ -260,27 +260,36 @@ func extractCosmosConfigDocs(ctx extractContext, r output.AzureResource, out *pi
...
+	maxDocSize := ctx.MaxCosmosDocSize
+	if maxDocSize <= 0 {
+		maxDocSize = defaultMaxCosmosDocSize
+	}
+	maxDocScan := ctx.MaxCosmosDocScan // 0 = unlimited
 
+	scanned := 0
 	for queryPager.More() {
+		if maxDocScan > 0 && scanned >= maxDocScan {
+			slog.Debug("cosmos doc scan limit reached", "db", dbName, "container", containerName, "limit", maxDocScan)
```
bernard-yip: slog.Info would be useful here while scanning imo

pkg/azure/extraction/extract_cosmos.go
```diff
@@ -260,27 +260,36 @@ func extractCosmosConfigDocs(ctx extractContext, r output.AzureResource, out *pi
...
+			return
+		}
+
 		page, err := queryPager.NextPage(ctx.Context)
```
bernard-yip, line 286: some progress logging might be useful to also see if Aurelian is stuck or is chugging along the large number of documents.

bernard-yip (commented):
Consider adding unit tests for the pkg components as well to ensure we handle the main/edge cases.

pkg/azure/extraction/extract_cosmos.go
ztgrace: bumped to info

pkg/azure/resourcegraph/lister.go
ztgrace: fixed

PR #133, opened 2026-03-21
Add IAM quick analyze module and multi-profile GAAD collection

Tanq16:
## Summary

- New `iam-quick-analyze` module for fast IAM recon — scans GAAD data for privilege escalation paths (39 combinations) and trust relationship misconfigurations without Neo4j or full resource enumeration
- Multi-profile support with parallel GAAD collection and merging across accounts
- Structured output: individual JSON files per privesc type and trust type, plus org hierarchy and per-account SCP mappings
- Offline analysis mode via `--gaad-dir` to re-analyze pre-collected GAAD files locally without AWS credentials
- Extended `account-auth-details` to support `--profiles` for multi-account GAAD collection, writing individual files to `{output-dir}/gaad/{accountID}.json`
- Exported `CollectOrgHierarchy` from orgpolicies package for reuse


ztgrace (commented):
# Review — IAM quick analyze + multi-profile GAAD

Reviewed against the `reviewing-aurelian-modules` skill and the `modern-golang` skill (project is on Go 1.25.3, so all modern patterns apply). Build is clean, unit tests pass.

## Blocker

### Integration test cannot pass against this `Run()`
`pkg/modules/aws/recon/iam_quick_analyze.go:60` — `Run()` calls `analyzer.Collect()`, which uses an internal pipeline and returns a `*Findings` struct. The findings are written to disk via `writeFindings` but **nothing is ever emitted to the `out *pipeline.P[model.AurelianModel]` parameter**.

`pkg/modules/aws/recon/iam_quick_analyze_integration_test.go:35-46` then does:
```go
pipeline.Pipe(p1, mod.Run, p2)
for m := range p2.Range() {
    if r, ok := m.(output.AWSResource); ok {
        findings = append(findings, r)
    }
}
require.NotEmpty(t, findings, "should produce at least some findings")
```
The test will always observe an empty `findings` slice. Either (a) have `Run()` emit each finding to `out` (in addition to the per-type files), or (b) change the test to read from the JSON output directory.

The cleanest fix is to drop `Collect()`, call `analyzer.Analyze(out)` to forward everything to the framework pipeline, and accumulate the per-type slices for disk writes inside `Run()` (or via a small consumer goroutine). That also removes the awkward `internal-pipeline + goroutine + Close + Range` dance in `Collect()`.

## High

### Hand-rolled ARN/principal parsing — flagged by the review skill
`trusts.go` and `analyzer.go` reimplement parsing four different ways:
- `extractAccountID` (analyzer.go:233) — `strings.SplitN(arn, ":", 6)`
- `extractAccount` (trusts.go:156) — `strings.Split`
- `extractServiceFromARN` (trusts.go:177) — `strings.Split(arn, "/")`
- `getPrincipalType` (trusts.go:186) — `strings.Split` + `strings.SplitN`

The skill explicitly flags these. Use `github.com/aws/aws-sdk-go-v2/aws/arn` (already used in `pkg/aws/cloudfront/lister.go:31`, `pkg/aws/enumeration/...`):
```go
import awsarn "github.com/aws/aws-sdk-go-v2/aws/arn"

parsed, err := awsarn.Parse(s)
if err != nil { return "" }
return parsed.AccountID
```
Bare account-ID principals (e.g. `"123456789012"`) and `"*"` aren't ARNs, so keep a small fallback for those, but use the SDK parser everywhere a real ARN is expected.

### Hierarchy is fetched twice
`collectOrg` (iam_quick_analyze.go:160-187) calls `CollectOrgPolicies` (which internally calls `collectOrganizationHierarchy`) and then immediately calls `CollectOrgHierarchy`, doing the full org tree walk twice. There's also dead code:
```go
hierarchyOpts := orgpolicies.CollectorOptions{...}
_ = hierarchyOpts // hierarchy comes from the same collection
```
Two ways to fix:
- Expose the hierarchy from `CollectOrgPolicies` (e.g. return both, or attach to `OrgPolicies`), and drop the second call. This is the right structural fix.
- At minimum, delete `hierarchyOpts` and its `_ =` line.

### `context.Background()` instead of `cfg.Context`
iam_quick_analyze.go:168, 182 use `context.Background()`. The sibling `org_policies.go:54` correctly uses `cfg.Context`. Pass through the framework context so cancellations propagate.

### Hierarchy error silently dropped
iam_quick_analyze.go:183-185:
```go
hierarchy, err := orgpolicies.CollectOrgHierarchy(context.Background(), opts)
if err != nil {
    return orgData, nil, nil // org data without hierarchy is still useful
}
```
Skill rule: `if err != nil { return nil }` without logging — must `slog.Warn` at minimum. Add `cfg.Log.Warn` (this function has access via the caller, so plumb it in).

## Modern Go violations

### `sync.WaitGroup` + index-based result array → `errgroup`
Both `iam_quick_analyze.go:281-306` (`collectGAADs`) and `account_auth_details.go:82-103` use the legacy `WaitGroup + results[idx]` pattern. The `modern-golang` skill flags this directly, and `pkg/modules/aws/recon/graph.go:88` in this same package already uses `errgroup`. Replace with:
```go
g, ctx := errgroup.WithContext(cfg.Context)
g.SetLimit(N) // bound concurrency to a sane default — avoid hammering AWS with 50+ profiles
results := make([]result, len(profiles))
for i, profile := range profiles {
    g.Go(func() error {
        ...
        results[i] = result{...}
        return nil
    })
}
if err := g.Wait(); err != nil { ... }
```
While you're there, drop the `go func(idx int, p string) {...}(i, profile)` arg-copy — Go 1.22+ has per-iteration loop scoping, the project is on 1.25.3.

### `%v` instead of `errors.Join` + `%w`
`iam_quick_analyze.go:343` and `account_auth_details.go:133`:
```go
return nil, nil, fmt.Errorf("all profiles failed: %v", errs)
```
Use `errors.Join`:
```go
return nil, nil, fmt.Errorf("all profiles failed: %w", errors.Join(errs...))
```

## Medium

### Unsafe type assertion in `Collect()`
`analyzer.go:57-61`:
```go
name := r.Properties["privesc_type"].(string)
```
If the value is missing or non-string this panics. Use the comma-ok form even though today's emitters all set strings — fragile contracts get broken later.

### Duplicated GAAD merge logic
`loadGAADsFromDir` (iam_quick_analyze.go:146-149) and `collectGAADs` (iam_quick_analyze.go:324-339) repeat the same four `Range`/`Set` calls plus the `multi-account` skeleton. Extract:
```go
func mergeGAAD(dst, src *types.AuthorizationAccountDetails) {
    src.Users.Range(func(arn string, u types.UserDetail) bool { dst.Users.Set(arn, u); return true })
    // ...
}
```
and a `func newCombinedGAAD() *types.AuthorizationAccountDetails`.

### Wrong `SupportedResourceTypes()`?
iam_quick_analyze.go:55-57 returns `AWS::Organizations::Account`. This module's actual input target is "one or more AWS profiles" and it scans the whole org policy tree from one of them — `org-policies.go` uses `AWS::Organizations::Organization` for the same shape. Worth aligning, and worth confirming what gateway/orchestrator code keys off of.

### Multi-account org collection silently uses only the first profile
iam_quick_analyze.go:165 — `Profile: m.Profiles[0]`. With multiple profiles passed in, only one account's view of the org is collected. If the first profile lacks `organizations:*` perms (common — payer-only), the entire org block is dropped. At minimum, document the requirement; better, try profiles in order until one succeeds.

## Minor

- `policyLookup` (analyzer.go:109-117) returns a closure when a plain method `(a *Analyzer) lookupPolicy(arn string) *types.Policy` works — pass `a.lookupPolicy` as the function value.
- Several doc comments on unexported helpers describe *what* the code does (e.g. analyzer.go:67, 86, 161, 184, 222, 232; trusts.go:150, 155, 164, 176, 185). The skill says to drop these unless behavior is genuinely surprising. The `actionGroup` doc (analyzer.go:120-126) is the kind worth keeping — that's a non-obvious invariant.
- `extractAccountID` argument name `arn string` is fine today, but if you adopt `aws/arn` as recommended above, alias the import (`awsarn "github.com/aws/aws-sdk-go-v2/aws/arn"`) so you don't shadow it. The cloudfront lister already does this.
- `writeJSON` (iam_quick_analyze.go:267) lives in the recon module — fine for now, but if a second module needs it, lift to a shared helper.

ztgrace (approved):

PR #156, opened 2026-04-07
Fix find-secrets memory leak: stream JSONL output, bump titus v1.1.19

ztgrace:
## Summary

- **Stream JSONL output** instead of accumulating all results in memory via `Collect()`. Results are written to disk one per line and freed immediately. Output extension changed from `.json` to `.jsonl`.
- **Switch `json.MarshalIndent` to `json.Marshal`** for proof bytes in `risk.go` — proof is embedded as a `[]byte` field, no need for pretty-printing.
- **Bump titus to v1.1.19** which fixes two critical memory issues in the matcher:
  - `runeSpanToByteSpan` was allocating ~6x content size per regex match (string copy + `[]rune` conversion + intermediate strings)
  - `regexp2.FindStringMatch` was converting content to `[]rune` **per rule** (~200 rules × 4x content size = 800x amplification). Fixed by converting once and using `FindRunesMatch`.

## pprof evidence

Before fixes — scanning a single AWS account:
- **920GB cumulative** from `regexp2.getRunes` (per-rule rune conversion)
- **128GB cumulative** from `runeSpanToByteSpan` (per-match rune conversion)  
- **17GB live** from `json.MarshalIndent` (result accumulation via `Collect()`)
- Process RSS: **44GB+**

After fixes — expected peak memory in low single-digit GB range.

## Test plan

- [ ] Run `find-secrets` against an AWS account and monitor RSS stays bounded
- [ ] Verify `.jsonl` output is valid (one JSON object per line, parseable by `jq`)
- [ ] Verify empty scans produce no output file (clean up behavior)
- [ ] Existing unit tests pass

### DIFF
--- a/cmd/generator.go
+++ b/cmd/generator.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"log/slog"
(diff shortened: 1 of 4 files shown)


xoverride:
Closing by PR 238, PR 239, PR 241

PR #158, opened 2026-04-08
aws recon find-secrets adding lambda env vars

ztgrace:
Adding the lambda properties as a scanning target.


ztgrace:
duplicated #155

PR #177, opened 2026-04-24
feat(fixture): destroy integration fixtures on package success

ztgrace:
## Summary

Flip the Aurelian integration-test fixture framework from "persist by default" to **"destroy on successful package run, persist on failure or opt-out."**

- **Default:** fixtures are torn down after every test package that passes. No more idle cloud cost between runs.
- **`--keep` / `AURELIAN_KEEP_FIXTURES=1`:** opt out of post-run destroy (local iteration).
- **`--redeploy` / `AURELIAN_REDEPLOY_FIXTURES=1`:** force teardown + redeploy at Setup (replaces the old `--destroy` / `AURELIAN_DESTROY_FIXTURES`, whose name collided with the new default).
- **Any test failure** in the package keeps all its fixtures alive for debugging.
- **Default `-timeout 30m`** applied automatically in `run-integration-tests.sh` so cleanup isn't SIGKILLed on long runs.

## Behavior matrix

| Env / flags | Setup | End-of-package |
|---|---|---|
| (none) | reuse if hash matches, deploy if not | **destroy iff package passed** |
| `--keep` / `AURELIAN_KEEP_FIXTURES=1` | same | skip destroy |
| `--redeploy` / `AURELIAN_REDEPLOY_FIXTURES=1` | force teardown + redeploy | destroy iff package passed |
| both | force teardown + redeploy | skip destroy |
| any test failed | — | skip destroy regardless |

## Implementation

- Process-level fixture registry in `test/testutil/fixture/` keyed by `StateKey`, mutex-guarded, dedupes shared fixtures across tests.
- `Setup()` registers in `globalRegistry` after successful deploy/reuse.
- New `fixture.RunTests(m *testing.M) int` — each integration-test package's `TestMain` calls it. On success (unless `AURELIAN_KEEP_FIXTURES=1`), drains the registry via `DestroyAll`.
- `DestroyAll` runs each fixture's teardown under a 10-minute per-fixture context; `RunTests` wraps the whole thing in a 15-minute outer bound. Errors are aggregated via `errors.Join` so one bad fixture doesn't block the rest.
- New `PurgeModulePrefix` S3 op sweeps the entire `integration-tests/<moduleDir>/` prefix post-destroy (empty state, hash marker, artifacts, `.tflock`) and surfaces `DeleteObjects` partial-failure errors.
- `teardownStack` extracted from `redeployStack` for reuse.
- TestMain added to 12 integration-test packages.
- `BaseFixture.logf` routes log output through `testing.T.Logf` during Setup and falls back to `log.Printf` afterward — fixes a panic where `teardownStack` tried to log to a completed test from the TestMain drain.

## Scripts

- `run-integration-tests.sh`: removed `--destroy`, added `--keep` and `--redeploy`, defaulted `-timeout 30m`, updated help + examples.
- `destroy-integration-fixtures.sh`: new bulk-cleanup tool for orphaned / SIGINT'd / debug-held fixtures. The preamble documents it as a complement to the now-automatic routine cleanup.

## Test plan

Validated end-to-end against the real integration state bucket:

- [x] Canary (`TestDestroyAll_RemovesModulePrefix`) provisions an S3 bucket, calls `DestroyAll`, asserts the S3 module prefix is empty after — passes.
- [x] `TestEC2IMDSEnricher` with default flags → fixture deployed, tests pass, TestMain drains, S3 prefix empty after.
- [x] Same test with `--keep` → S3 prefix populated after (destroy skipped).
- [x] Same test with `--redeploy --keep` → log shows `teardown + redeploy (forced)` even on hash match; S3 populated after.
- [x] Same test with default flags again → reuses fixture, passes, TestMain drains, S3 prefix empty after.
- [x] `./scripts/run-integration-tests.sh --destroy` → exits with `Unknown argument: --destroy`.
- [x] All 17 fixture package unit tests pass (registry, RunTests decision matrix, lifecycle branches, env var regression).
- [x] `go vet -tags=integration` clean across every package that gained a TestMain.

Failure-path (any test fail → keep fixtures) is covered by `TestRunTestsWith_TestsFailed_SkipsDestroy` in the unit suite.

### DIFF
--- a/pkg/aws/enumeration/testmain_test.go
+++ b/pkg/aws/enumeration/testmain_test.go
@@ -1,0 +1,12 @@
+//go:build integration
+
+package enumeration
+
+import (
+	"os"
+	"testing"
+
+	"github.com/praetorian-inc/aurelian/test/testutil/fixture"
+)
+
+func TestMain(m *testing.M) { os.Exit(fixture.RunTests(m)) }
--- a/pkg/aws/gaad/testmain_test.go
+++ b/pkg/aws/gaad/testmain_test.go
@@ -1,0 +1,12 @@
+//go:build integration
+
+package gaad
+
+import (
+	"os"
+	"testing"
+
+	"github.com/praetorian-inc/aurelian/test/testutil/fixture"
+)
+
+func TestMain(m *testing.M) { os.Exit(fixture.RunTests(m)) }
--- a/pkg/aws/resourcepolicies/testmain_test.go
+++ b/pkg/aws/resourcepolicies/testmain_test.go
@@ -1,0 +1,12 @@
+//go:build integration
+
+package resourcepolicies
+
+import (
+	"os"
+	"testing"
+
+	"github.com/praetorian-inc/aurelian/test/testutil/fixture"
+)
+
+func TestMain(m *testing.M) { os.Exit(fixture.RunTests(m)) }
--- a/pkg/azure/extraction/testmain_test.go
+++ b/pkg/azure/extraction/testmain_test.go
@@ -1,0 +1,12 @@
+//go:build integration
+
+package extraction
+
+import (
+	"os"
+	"testing"
+
+	"github.com/praetorian-inc/aurelian/test/testutil/fixture"
+)
+
+func TestMain(m *testing.M) { os.Exit(fixture.RunTests(m)) }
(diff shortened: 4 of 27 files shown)


xoverride (changes requested):
See [PR237](https://github.com/praetorian-inc/aurelian/pull/237)

PR #192, opened 2026-04-27
list-all: derive resource type coverage from registered modules

ztgrace:
## Summary

Replaces the hand-curated `pkg/aws/resourcetypes` `all` slice with a runtime-derived value computed once per process from `baseline ∪ registered-AWS-modules.SupportedResourceTypes() − exclusions`. Consumer modules (find-secrets, public-resources, resource-policies, ...) are now the source of truth for what `list-all` enumerates; new types declared by a module are auto-included with no manual sync.

- **Plugin registry:** new `plugin.ByPlatform(p Platform) []Module` iterator (used by the union builder).
- **Exclusions:** new `pkg/aws/resourcetypes/exclusions.go` with `IsExcluded()` predicate and a justification-required map. Initial entry: `AWS::Organizations::Account` (Guard pseudo-input — listing it would discover sibling accounts in the org).
- **Runtime union:** new `pkg/aws/resourcetypes/union.go` with `sync.Once`-cached `computeAll()`; validates each declared type against `^AWS::[A-Z][A-Za-z0-9]*::[A-Z][A-Za-z0-9]*$` and `slog.Warn`-drops malformed entries.
- **`coverage_test.go`** (external test pkg, no build tag): blank-imports the loader so the registry is populated; headline `TestModuleConsumerCoverage` is the drift gate. The integration build tag was deliberately not used because, per project convention, that tag is reserved for tests that deploy real infra.

The headline drift test surfaced 7 spec-named missing types plus 4 additional cross-module types (`AWS::EC2::Image`, `AWS::SSM::Document`, `AWS::Cognito::UserPool`, `AWS::EFS::FileSystem`, `AWS::Redshift::Cluster`, `AWS::OpenSearchService::Domain`, `AWS::Elasticsearch::Domain`, `AWS::Organizations::Organization`, `AWS::IAM::Group`, ...) that the old hand-curated list missed. All now flow through the union.

## Test Plan

- [x] `go test ./pkg/plugin/... -run TestByPlatform` — 3/3 pass
- [x] `go test ./pkg/aws/resourcetypes/...` — 22/22 pass (14 internal + 8 external)
- [x] `go test ./pkg/aws/resourcetypes/... -run TestModuleConsumerCoverage` — drift gate green
- [x] `go test ./...` — full repo green
- [x] `go test -race ./pkg/aws/resourcetypes/...` — race-clean (verified during review)
- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] Manual smoke: run \`aurelian aws recon list-all --scan-type full --regions us-east-1\` against an account with AMIs / EFS / Cognito and confirm the previously-missing types appear in output.

## Reviewer Follow-Ups (Optional, Non-Blocking)

Code-quality review surfaced a few recommendations the plan deliberately deferred:

- **Refactor `IsValid` / `Validate` / `computeAll` ergonomics.** The pattern \`_ = computeAll() // ensure allSet is populated\` could be cleaner as an \`ensureComputed()\` helper. The plan kept the explicit form on purpose, but a follow-up that introduces \`ensureComputed()\` would also let \`Validate\` hoist the call out of its loop.
- **Document cache lifetime.** \`computeAll()\` is process-lifetime cached and is **not** invalidated by \`plugin.ResetRegistry()\`. No tests rely on that today, but a one-line note in the package doc-comment would prevent future debugging pain.
- **Move \`TestExclusions_AreReferenced\` internal.** Currently in \`coverage_test.go\` (external pkg) with a hardcoded \`knownExclusions\` literal because \`exclusions\` is unexported. Moving the test into \`types_test.go\` (internal) would eliminate the silent-skip risk if someone adds an exclusion and forgets to mirror it here. File-move only, no API change.

Plan and spec checked in under \`docs/superpowers/\` on the prior branch (\`zg/lab-2525-enhance-listing-capabilities-in-aws\`).

### DIFF
--- a/pkg/aws/resourcetypes/coverage_test.go
+++ b/pkg/aws/resourcetypes/coverage_test.go
@@ -1,0 +1,38 @@
+// Package resourcetypes_test is an external test package that imports the
+// module loader so the plugin registry is fully populated before tests run.
+// Tests here observe the runtime resource-type union; static-data tests live
+// in types_test.go (internal package).
+package resourcetypes_test
+
+import (
+	"testing"
+
+	"github.com/praetorian-inc/aurelian/pkg/aws/resourcetypes"
+	"github.com/praetorian-inc/aurelian/pkg/plugin"
+
+	// Blank-import the loader to register all modules before tests run.
+	_ "github.com/praetorian-inc/aurelian/pkg/modules/loader"
+)
+
+// TestModuleConsumerCoverage is the headline drift test: every type a
+// registered AWS module declares in SupportedResourceTypes() must be in
+// GetAll() or in the exclusions list. Failing this test means a consumer
+// module added a type that list-all won't enumerate — fix by either adding
+// the type to baseline (if listing makes sense) or to exclusions (with
+// justification).
+func TestModuleConsumerCoverage(t *testing.T) {
+	all := make(map[string]bool)
+	for _, rt := range resourcetypes.GetAll() {
+		all[rt] = true
+	}
+
+	for _, m := range plugin.ByPlatform(plugin.PlatformAWS) {
+		for _, rt := range m.SupportedResourceTypes() {
+			if all[rt] || resourcetypes.IsExcluded(rt) {
+				continue
+			}
+			t.Errorf("module %q declares %q which is not in GetAll() and not excluded",
+				m.ID(), rt)
+		}
+	}
+}
--- a/pkg/aws/resourcetypes/coverage_test.go
+++ b/pkg/aws/resourcetypes/coverage_test.go
@@ -109,8 +109,6 @@ func TestExclusions_AreReferenced(t *testing.T) {
 	// weight from a past refactor — delete it.
 	referenced := make(map[string]bool)
 
-	// Baseline references — re-derive from GetAll() ∪ exclusions to avoid
-	// reaching into the package's unexported baseline slice.
 	for _, rt := range resourcetypes.GetAll() {
 		referenced[rt] = true
 	}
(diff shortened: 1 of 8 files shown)


ztgrace:
Addressed reviewer feedback in `5a669b41`:

**Critical (Gemini, Claude): unresettable cache → test pollution**
Added `ResetForTest()` in a new `export_test.go` (visible only to test binaries via the `_test.go` suffix). Tests that mutate the plugin registry can now invalidate the cache before re-fetching. Added `TestResetForTest_AllowsRecomputation` to lock in the contract.

**Critical (Claude): `IsValid` side-effect pattern**
Removed `computeAll()` entirely. Replaced with `ensureComputed()` which returns nothing — `IsValid` and `GetAll` now call it explicitly, then read `allSet`/`allCache` directly. `Validate` hoists the `ensureComputed()` call out of its loop and reads `allSet` per element rather than chaining through `IsValid`. The `_ = computeAll() // ensure allSet is populated` line is gone.

**Coverage gap (Gemini): hardcoded `knownExclusions`**
Added `ExclusionsForTest()` to `export_test.go` exposing the keys of the unexported `exclusions` map. `TestExclusions_AreReferenced` now iterates real exclusions; new entries are auto-validated for "must be referenced by baseline or a module."

**Init-order risk (Gemini Critical, descriptive)**
Added doc-comment paragraphs in both `types.go` (package doc) and `union.go` (on `ensureComputed`) noting:
- Cache is process-lifetime; not invalidated by `plugin.ResetRegistry()`.
- Don't call `GetAll`/`IsValid`/`Validate` from package `init()` — runs before all module init() and would permanently freeze a partial union.
- Tests that need a fresh union should call `ResetForTest`.

**Not addressing: dynamic `summary` filtering (Gemini suggestion)**
The design intentionally keeps `summary` static. `TestSummary_SubsetOfGetAll` already enforces the invariant; if a custom build excludes modules that `summary` references, the test should fail loudly rather than silently filter.

Test summary after the fix: 23 tests in `pkg/aws/resourcetypes/...` all pass; full repo `go test ./...` clean; `go vet` and `gofmt` clean.

Tanq16 (approved):

PR #194, opened 2026-04-30
Add Azure APIM Audit module (missing-auth + backend-direct-access)

silentmode19:
## Summary

Adds `azure/recon/apim-audit` — a read-only ARM-authenticated recon module that audits every APIM service in scope across two checks, reusing one Resource Graph discovery pass.

- **Missing-auth check** — for each API (including native MCP-type APIs and proxy-shaped MCPs), inspects inbound policy XML at the service / product / API scopes for `validate-jwt`, `validate-azure-ad-token`, `ip-filter`, and `check-header`, and combines with the API's `subscriptionRequired` flag. APIs with no auth at any scope emit `azure-apim-missing-auth` (Critical), or `azure-apim-mcp-missing-auth` (Critical) when the API is MCP-shaped.
- **Backend direct-access check** — lists each service's configured backends, classifies the URL by hostname suffix (Azure App Service / ASE / chained APIM / GCP Cloud Run / OpenShift / other), and for App Service backends queries Azure Resource Graph for `publicNetworkAccess` and IP restriction count. Reachable backends emit `azure-apim-backend-direct-access` (High); non-correlated and non-Azure backends emit `azure-apim-backend-unverified` (Low) for human triage.
- Native MCP-type APIs surfaced via the `2024-06-01-preview` API version (the GA listing hides them).
- Module ID `apim-audit`. `OpsecLevel = moderate`. No active probing — every signal comes from the ARM control plane or Azure Resource Graph.

## Risk Shapes

| Name | Severity | Trigger |
|---|---|---|
| `azure-apim-missing-auth` | Critical | API has no auth at service / product / API scope and `subscriptionRequired = false` |
| `azure-apim-mcp-missing-auth` | Critical | Same as above, but API is detected as an MCP server (native type or `/mcp`/`/sse`/`/messages` operations) |
| `azure-apim-backend-direct-access` | High | App Service backend with `publicNetworkAccess = Enabled` and ≤ 1 IP restriction rule (the default deny-all counts as none) |
| `azure-apim-backend-unverified` | Low | Backend hostname is non-Azure or did not match any App Service in accessible subscriptions — flag for manual review |

`DeduplicationID` is `<apim_service_resource_id>/<api_id>` (missing-auth) or `<apim_service_resource_id>/backend/<backend_name>` (backend). `ImpactedResourceID` is always the APIM service.

## Architecture Notes

- Discovery (`subscriptions → ARG → Microsoft.ApiManagement/service → apimServiceTarget`) lives once at the top of `Run()`. Each target fans through every registered `apimCheck` in a single goroutine, sharing one `armapimanagement.ClientFactory` and one `armresourcegraph.Client`. Adding a third check is one struct member.
- Reusable APIM helpers under `pkg/azure/apim/`: `list.go` (paginated API listing), `inventory.go` (MCP detection + verdict shaping), `policy.go` (inbound policy XML parsing), `backend.go` (URL → category classification). Each is independently unit-tested and would also be reusable from a future module if needed.
- Earlier WIP commits (`c3d565a6`, `8d0c29f7`, `c13488fd`, `7f4ebecc`) introduced two separate modules (`apim-missing-auth`, `apim-backend-direct-access`); the consolidation commit (`9be90fad`) merges them into one module with a checker registry, since they share input parameters, ARM client, opsec level, and discovery logic. The CLI exposes a single `apim-audit` ID — the older IDs are gone.

## Test Plan

- [x] `go build ./...` clean
- [x] `go vet ./pkg/azure/apim/... ./pkg/modules/azure/recon/...` clean
- [x] `go test ./pkg/azure/apim/... ./pkg/modules/azure/recon/...` passes (unit tests for policy parsing, MCP detection, App Service exposure heuristic, captured ARM fixture for native MCP listing)
- [x] `go run . list-modules | grep apim-audit` confirms registration; old `apim-missing-auth` / `apim-backend-direct-access` IDs are gone
- [x] **Live integration test** — `test/terraform/azure/recon/apim-audit/` fixture provisions 2 APIM Consumption services + 7 APIs + 1 public-reachable App Service backend; `go test -tags=integration -timeout 30m -run '^TestAzureAPIMAudit$' ./pkg/modules/azure/recon/` passes 17/17 subtests against live Azure
- [x] Positive: unauth-api flagged Critical, fake-mcp-api flagged Critical with `is_mcp_server=true`, public App Service backend flagged High with `publicNetworkAccess=Enabled` and ARG-correlated `correlated_resource_id`
- [x] Negative: jwt-api, ipfilter-api, checkheader-api, product-auth-api, inherits-auth-api (apim2, service-scope JWT) correctly NOT flagged
- [x] Invariants: no duplicate `DeduplicationID`s, all risk names within documented set, all impacted IDs reference an APIM service

## Out of Scope

- Active HTTP probing of backends (separate triage skill — see discussion in commit log)
- Triage / severity-upgrade logic for `azure-apim-backend-unverified` (will live in a Claude triage skill that consumes this module's JSON output)
- APIM developer portal cross-tenant signup attack (separate module — see #161)


silentmode19:
Addressed in 3 fix commits on top of the branch.

**Critical correctness — fixed**

- **Policy inheritance bypass.** `AuthPosture` now records `HasBase`, the parser detects `<base />` inside `<inbound>`, and `ClassifyAPI` only counts service-scope auth when `api.APIPolicyAuth.HasBase` is true. Added `apim.InheritedFromParent()` for callers to model "no custom policy at this scope" as the implicit base-chain APIM applies; the missing-auth fetcher now uses it on 404 responses. Tests cover the bypass cases (commit `184f4545`).
- **Weak-product bypass.** `ClassifyAPI` now requires *every* associated product to authenticate before reporting product-scope auth. A product authenticates if it has its own auth, or if `<base />` chains to a service-scope auth. Test `TestClassifyAPI_ProductPolicyAuthRejectsAnyOpenProduct` covers the multi-product attacker-uses-weakest-key case (commit `184f4545`).
- **Pagination retry.** `ListAPIs` now wraps non-2xx ARM responses in `runtime.NewResponseError(resp)` so the standard Azure paginator's retry decision (which uses `errors.As` against `*azcore.ResponseError`) actually matches and retries 429/503 (commit `ce403de3`).

**Security — fixed**

- **Kusto injection via FQDN.** `lookupAppServiceExposure` now escapes `\` to `\\` before escaping `'` to `''`. The order matters because `\` would otherwise consume the closing quote (commit `5db70b0a`).
- **NextLink token leakage.** `ListAPIs` now validates that the `nextLink` URL is HTTPS and its host matches the original ARM endpoint before attaching the bearer token. Foreign hosts, subdomain-attack hostnames, and non-HTTPS values are all rejected. Test matrix covers the validation cases (commit `ce403de3`).

**Suggestions — fixed**

- **MCP trailing slash.** `IsMCPServer` now trims trailing slashes before suffix-matching, so `/mcp/` and `/v1/mcp/` are detected (commit `184f4545`).
- **Nil ARG client noise.** `classifyAndEmitBackend` now skips App Service correlation entirely when `argClient` is nil (the audit module already logs a warning at startup if ARG init fails). Previously every Azure App Service backend produced a Low `azure-apim-backend-unverified` finding when ARG was unavailable (commit `5db70b0a`).

**Deferring — operation-level auth fallback**

The reviewer used "consider", and the lift is real: `Operations` doesn't currently carry policy data at all (`apim_audit_missing_auth_check.go:listAPIOperations` only fetches URL template fields), so implementing the rollup means N additional `factory.NewAPIOperationPolicyClient().Get` calls per API per service plus a caching/cost decision. The current limitation is acknowledged at `inventory.go` already. Tracking as a follow-up — happy to open an issue if you'd like a placeholder.

Build + vet + unit tests all green:

```
ok  	github.com/praetorian-inc/aurelian/pkg/azure/apim	1.458s
ok  	github.com/praetorian-inc/aurelian/pkg/modules/azure/recon	0.816s
```

silentmode19:
Round 2 — addressed the second human review and CodeRabbit's 6 inline + nit comments. 4 commits on top.

**Critical correctness — fixed**

- **`isNotFound` conflated 404 with 401/403.** Permission-denied reads were silently downgraded to "no auth", producing false-positive Critical findings when the auditor lacked policy read on a scope. Split into `isPolicyNotConfigured` (404 → `InheritedFromParent`) and propagate other errors. The three policy fetchers now return `(AuthPosture, error)` and `checkMissingAuth` skips the API with a warning on any read failure, rather than reporting unauthenticated based on partial data (commit `30e6842a`).
- **`IncludeFragment` short-circuit treated unknown fragments as authenticating.** Every callsite passes `nil` for `fragmentAuth`, so any policy with `<include-fragment>` (logging, header transforms, CORS) was being classified as authenticated — a silent false-negative on missing-auth detection. Drop the `!known` short-circuit; unknown fragments now contribute nothing. Pre-indexing fragments via the policy-fragment client is a follow-up (commit `30e6842a`).
- **`openShiftRoutePattern` matched `.apps.` as a substring** → false positives like `web.apps.contoso.com` mislabeled as openshift. Drop the openshift category entirely; affected hosts fall through to "other" and still emit `azure-apim-backend-unverified` Low for manual triage. Test fixture updated to assert the regression case (commit `9b4ca9f3`).

**Security / reliability — fixed**

- **`http.DefaultClient` had no per-request timeout.** Hung ARM endpoints could tie up worker goroutines under context-less callers. Use a dedicated `*http.Client` with `Timeout: 30s` (commit `0d7e90b5`).

**CodeRabbit specific**

- **MCP detection too liberal.** Lone `/messages` (or `/sse`) is ambiguous — many ordinary REST APIs serve those paths. Now require the SSE pair (`/sse` + `/messages` or `/message`); `/mcp` standalone is still sufficient (Streamable HTTP transport). Tests cover the lone-`/messages`, lone-`/sse`, lone-`/message` regression cases (commit `9c1972a9`).
- **ARG query missed App Service deployment slots.** Slots are a separate resource type (`microsoft.web/sites/slots`) with their own hostnames like `<app>-staging.azurewebsites.net`. Backends pointing at slot URLs missed correlation and were downgraded to backend-unverified. Extend the `where` clause (commit `9b4ca9f3`).
- **`checkBackendDirectAccess` ran with `argClient = nil`.** Now gated at the audit dispatcher: when ARG init fails, the check is dropped from the dispatch slice instead of running anyway. Cleaner than the per-category nil guard inside the check (commit `30e6842a`/`9b4ca9f3`).
- **Protocols comparison length-only.** Use `reflect.DeepEqual` (commit `0d7e90b5`).

**Skipped / not applicable**

- Multi-product bypass — already addressed in `dcbaf1cc` from round 1; CodeRabbit's status footer confirms.
- Skip product-policy fetch when `SubscriptionRequired = false` — pure cost optimization not correctness, and the additional `ProductPolicyAuths` payload is harmless when `SubscriptionRequired` is false (the classifier ignores them in that case). Worth a follow-up issue if API-call cost becomes a concern.

**Verified**

```
go build ./...         ✓
go vet ./pkg/azure/apim/... ./pkg/modules/azure/recon/...  ✓
go test ./pkg/azure/apim/... ./pkg/modules/azure/recon/...
ok      github.com/praetorian-inc/aurelian/pkg/azure/apim       0.704s
ok      github.com/praetorian-inc/aurelian/pkg/modules/azure/recon 0.817s
go test -tags=integration -count=1 -run='^XYZ_NOMATCH$' ./pkg/modules/azure/recon/...  ✓ (compile-check only — live integration test was passed in round 1)
```

I have not re-run the live integration test against Azure since the test fixture was torn down. Happy to redeploy and re-run if you want pre-merge confirmation.

Elgin-Lee (approved):

PR #197, opened 2026-05-13
fix(aws/recon): continue enumeration on per-region AccessDenied/SCP errors

bernard-yip:
## Summary

- AWS recon previously halted on the first `AccessDeniedException` (e.g. an SCP `explicit deny` on a single region/service), making the tool fragile against partitioned, SCP-restricted AWS environments
- Adds a shared smithy-code-based `IsSkippableAWSError` classifier + thread-safe `SkipReport` aggregator, wired through every native enumerator (Amplify, S3, EC2, SSM, IAM) and CloudControl
- The `list-all` module now emits a grouped skip summary via `defer` so operators see the skipped `(region, service)` pairs on both success and partial-failure paths

Skippable error codes: `AccessDenied`, `AccessDeniedException`, `UnauthorizedOperation`, `AuthFailure`, `OptInRequired`, `InvalidClientTokenId`, `UnrecognizedClientException`, `TypeNotFoundException`, `UnsupportedActionException`, plus DNS/endpoint-not-found (region unsupported).

Fatal errors (STS `GetAccountID`, AWS config setup, no configured regions) remain fatal.

Closes [LAB-3328](https://linear.app/praetorianlabs/issue/LAB-3328)

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go vet -tags=integration ./...`
- [x] `go test ./pkg/aws/enumeration/... ./pkg/modules/aws/recon/...`
- [x] `go test -race ./pkg/aws/enumeration/...`
- [x] New unit tests in `pkg/aws/enumeration/continue_on_denied_test.go` (9 cases) — verify the loop continues across regions when one returns `AccessDeniedException`, and that `SkipReport` is populated with the correct tuple
- [x] Full real-AWS integration suite (`AWS_PROFILE=bfr go test -tags=integration -timeout 30m ./pkg/modules/aws/recon/...`): `TestAWSList` 7/7 sub-tests PASS in 28s; logs confirm the fix engaging on real AWS — `WARN skipping cloudcontrol ListResources type=AWS::OpenSearchService::Domain code=UnsupportedActionException` and the run continued through downstream tests
- [x] Two pre-existing failures (`TestAWSFindSecrets/detects_secret_in_ECS_Task_Def`, `TestAWSIAMQuickAnalyze`) reproduced on `main` (commit `ecccd05f`) — unrelated to this fix

### DIFF
--- a/pkg/aws/enumeration/amplify_enumerator.go
+++ b/pkg/aws/enumeration/amplify_enumerator.go
@@ -19,13 +19,15 @@ import (
 // because CloudControl does not support AWS::Amplify::App.
 type AmplifyAppEnumerator struct {
 	plugin.AWSCommonRecon
-	provider *AWSConfigProvider
+	provider   *AWSConfigProvider
+	skipReport *SkipReport
 }
 
-func NewAmplifyAppEnumerator(opts plugin.AWSCommonRecon, provider *AWSConfigProvider) *AmplifyAppEnumerator {
+func NewAmplifyAppEnumerator(opts plugin.AWSCommonRecon, provider *AWSConfigProvider, skipReport *SkipReport) *AmplifyAppEnumerator {
 	return &AmplifyAppEnumerator{
 		AWSCommonRecon: opts,
 		provider:       provider,
+		skipReport:     skipReport,
 	}
 }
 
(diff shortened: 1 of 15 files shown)


bernard-yip:
closing the PR to work on #200 instead.

PR #200, opened 2026-05-29
fix(aws/enumeration): resilient enumeration with smithy-based error classifier

xoverride:
## Summary

AWS recon (`list-all`) aborted on the first `AccessDeniedException` from any region/service (e.g. SCP explicit deny), abandoning all remaining work. This PR makes enumeration resilient: errors are classified, recorded, and skipped so the pipeline continues, while fatal credential errors still propagate.

### Verified end-to-end

| Scenario | Resources | Skipped | Abort? |
|---|---|---|---|
| Full access, full scan, 3 regions | 177 | 6 (natural CC unsupported) | No |
| Restricted role, full scan, 3 regions | 149 | 84 | **No** (was: abort at first error) |
| Restricted role, summary scan, 3 regions | 91 | 18 | No |

28 resources lost = exactly the types denied by the restricted role. All allowed types (S3, IAM, EC2, Lambda) have identical counts between full and restricted access.

### Architecture

- **Two-layer error handling**: inner loops use `ClassifySkippable` + `RecordBatch` for per-region granularity; dispatcher `handleListError` is a safety net so new enumerators get resilience by default
- **Blacklist classifier**: all AWS API errors are skippable except fatal credential errors (`ExpiredToken`, `SignatureDoesNotMatch`, etc.). This handles `GeneralServiceException` (CloudControl wrapping AccessDenied), `ThrottlingException`, `InvalidRequestException`, and every other service-level error
- **Lock-free hot path**: `ClassifySkippable` acquires no locks — goroutines accumulate skips locally and flush via `RecordBatch` once per region, O(regions) not O(errors)
- **`*SkipReport` threaded through constructors**: compile-time enforcement — missing param won't compile
- **`defer lister.Close()`**: logs grouped summary + writes `enumeration-skips.json` detail file. `sync.Once` prevents double-log
- **Compact summary**: region lists collapse to "N regions" when >5; all error codes shown with counts; per-skip logs at Debug, summary at Warn

### Operator experience

Before:
```
[*] enumerating 41 resource types across 3 regions
Error: list AWS::ECS::Cluster: GeneralServiceException: AccessDeniedException...
```

After:
```
[*] enumerating 41 resource types across 3 regions
[+] enumerated 149 resources
[!] skipped 84 operations across 3 regions:
  amplify ListApps: us-east-1, us-east-2, us-west-2 [AccessDeniedException×3]
  cloudcontrol ListResources: us-east-1, us-east-2, us-west-2 [AccessDeniedException×45, GeneralServiceException×18, InvalidRequestException×9, UnsupportedActionException×6]
  ssm ListDocuments: us-east-1, us-east-2, us-west-2 [AccessDeniedException×3]
[+] output written to aurelian-output/list-all-20260601.json
```

### AST enforcement (7 checks, 0.1s, no AWS needed)

| # | Test | Prevents |
|---|---|---|
| 1 | `TestNewEnumeratorRequiresClose` | Missing `defer Close()` — tied to the specific enumerator variable, not any unrelated Close |
| 2 | `TestSkipReportFieldRequiresUsage` | `skipReport` field without ClassifySkippable + `*op` deref + RecordBatch |
| 3 | `TestClassifySkippableServiceNames` | CC type string `"AWS::..."` as service name literal |
| 4 | `TestNewEnumeratorSharesSkipReport` | Non-shared SkipReport instance (checks variable identity) |
| 5 | `TestClassifySkippableAlwaysInIfInit` | Bare ClassifySkippable call without nil check |
| 6 | `TestFatalErrorCodes_AllBlocked` | New fatal code without test |
| 7 | `TestEnumerateByARN_DispatcherExtractsServiceAndRegion` | Dispatcher passing raw ARN instead of parsed service/region |

### Integration tests (32 sub-tests, 4 IAM roles, 3 regions, ~40 fixture resources)

Terraform deploys across `us-east-2`, `us-east-1`, `us-west-2` with 4 restricted IAM roles:
- **restricted**: S3/IAM/EC2/Lambda allowed, Amplify/SSM denied
- **wiring-check**: ALL services denied — verifies every enumerator's inner ClassifySkippable produces correct short service name
- **region_restricted**: mosaic deny (Amplify in us-east-1, Lambda in us-east-2)
- **partial_ec2**: DescribeImages allowed, DescribeImageAttribute denied

All regions, role ARNs, resource ARNs, and deny mappings from Terraform outputs — zero hardcoded values.

| Test | Error class | Assertion |
|---|---|---|
| S3/IAM/Lambda/EC2 allowed | None | Every fixture resource by ARN |
| Amplify/SSM denied | Full service deny | Exactly 0 results |
| MixedPipeline (8 types, interleaved) | Mixed allow/deny | All 40+ fixture resources by ARN survive |
| WiringCheck (5 types) | All services denied | Every enumerator produces exact short service name, no AWS:: prefix |
| PerRegionDenial (mosaic) | Different services in different regions | Per-region ARN presence/absence |
| MosaicPipeline (4 types × 3 regions) | Mosaic deny | Lambda us-east-1+us-west-2 survive, us-east-2 absent by ARN |
| PartialEC2Access | DescribeImageAttribute denied | 0 enriched (with precondition), no abort |
| ResourcePolicyDeny | S3 bucket policy deny | By-ARN skipped, ListBuckets still returns bucket |
| ByARN mixed allow/deny | IAM ARN + Amplify ARN | IAM present, Amplify absent |
| CloseWritesDetailFile | Real AWS errors | JSON file with AccessDenied entries |

### Related

- PR #178 (ztgrace) — similar fix with handleListError + Config fallback
- PR #197 (bernard-yip) — similar fix with IsSkippableAWSError + SkipReport
- LAB-3937 — follow-up: enrichment layer resilience

Closes LAB-2525


xoverride:
### Code review

Found 3 issues:

1. **`IsSkippableAWSError` blocklist design silently swallows `ThrottlingException` and other transient errors** — The function classifies ALL smithy errors as skippable unless they match one of 6 fatal credential codes. This means `ThrottlingException`, `ServiceUnavailableException`, and `InternalServerError` are silently recorded as skips with no retry. A previous commit (`9562f79c`) had a test `TestContinueOnDenied_NonListedSmithyCode_PropagatesFatal` that explicitly guarded against this by requiring `ThrottlingException` to propagate as fatal. That test was replaced with one testing `ExpiredToken` instead, removing the guard. Under throttling, an entire region's enumeration results can be silently dropped.

https://github.com/praetorian-inc/aurelian/blob/48cbd456cd5be9517a02f33d202a11fefffe8493/pkg/aws/enumeration/skiperror.go#L12-L35

2. **`IAMEnumerator.skipReport` is dead code — IAM denials propagate as fatal errors, breaking the resilience contract** — Every other enumerator (amplify, s3, ec2, ssm, cloudcontrol) calls `ClassifySkippable` + `RecordBatch` in their inner loops. `IAMEnumerator` accepts a `*SkipReport` in its constructor and stores it, but `listRoles`, `listPolicies`, and `listUsers` all return raw errors directly (e.g., `return false, fmt.Errorf("list IAM roles: %w", err)`) without calling `ClassifySkippable`. The `skipReport` field is never referenced after construction. An SCP denying `iam:ListRoles` will propagate as a hard error and cancel the errgroup, potentially aborting the entire enumeration run — the exact failure mode this PR is supposed to prevent.

https://github.com/praetorian-inc/aurelian/blob/48cbd456cd5be9517a02f33d202a11fefffe8493/pkg/aws/enumeration/iam_enumerator.go#L22-L58

3. **`handleListError` passes the full ARN as the `service` parameter, producing high-cardinality skip records with empty region** — On the ARN enumeration path, `identifier` is the full ARN string (e.g., `arn:aws:s3:::my-bucket`). When `handleListError` calls `ClassifySkippable(err, identifier, "List", "")`, the full ARN becomes the `Service` field and region is always empty. The parsed ARN (with `.Service` and `.Region`) is already available at the call site (line 107) but is not used. This breaks the skip report's per-service/per-region grouping, making the operator summary cluttered and ungrouped.

https://github.com/praetorian-inc/aurelian/blob/48cbd456cd5be9517a02f33d202a11fefffe8493/pkg/aws/enumeration/enumerator.go#L106-L131

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

xoverride:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Previous review findings (ThrottlingException classification, IAM skipReport dead code, ARN-as-service in handleListError) have all been addressed in subsequent commits (b0572c7b, b0b98777, 3d712a03).

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

bernard-yip:
Hey @xoverride , could you fix the merge conflicts and run `gofmt -w pkg/aws/enumeration/` before i do another review for any critical issues?

I'll most likely run pr-review from the skills plugin as a first pass to identify any issues.

pkg/aws/enumeration/skiperror_fatal_test_mode.go
```diff
@@ -0,0 +1,10 @@
+//go:build fatal_test_mode
+
+package enumeration
+
+func init() {
+	// For binary-level testing: make AccessDeniedException fatal so that
+	// the restricted role (which triggers real AccessDeniedException at the
+	// enumeration level) causes the pipeline to abort.
+	fatalErrorCodes["AccessDeniedException"] = struct{}{}
+}
```
bernard-yip, line 10: I dont see any tests here. is this expected?

pkg/aws/enumeration/fatal_error_binary_test.go
```diff
@@ -0,0 +1,119 @@
...
+		"--profile-dir", profileDir,
+		"--regions", "us-east-1",
+		"--scan-type", "summary",
+		"--output-dir", outputDir,
+		"--no-color",
+	)
+	cmd.Env = append(os.Environ(),
+		"AWS_CONFIG_FILE="+filepath.Join(profileDir, "config"),
+		"AWS_SHARED_CREDENTIALS_FILE="+filepath.Join(profileDir, "credentials"),
+	)
```
bernard-yip: is it necessary for us to run binaries where it reads our .aws/{config,credentials} files for the fatal errors?

I was hoping it could be part of the integration test rather than a binary test.

pkg/aws/enumeration/skiperror_fatal_test_mode.go
xoverride: This is expected as it's using as build tag

pkg/aws/enumeration/fatal_error_binary_test.go
xoverride: Agree; this is part of my local binary test that I run for dev and validation for a larger test suite; should not be used as part of standard CI testing as is.

bernard-yip: In that case I'd rather drop the binary build from the PR entirely.

We can verify the fatal path in the integration test instead: insert a common error (e.g. `AccessDeniedException`) into `fatalErrorCodes` via a test seam and assert the pipeline aborts via the e2e tests.

Basically what you're already doing in `pkg/aws/enumeration/skiperror_fatal_test_mode.go`, just in-process instead of through the binary. That covers the "do we actually exit on a fatal code" case without the harness.

Any thoughts on this?

pkg/aws/enumeration/z3_stress_enumerator_test.go
```diff
@@ -0,0 +1,451 @@
...
+// - deep fmt.Errorf wrapping chains before classification
+// - errors returned from a closure and classified by the caller
+//
+// "Bug" patterns (stressBug_) must be detected as violations:
+// - silent drop (count=0)
+// - double classify (count=2)
+// - classify then leak (missing return, count=2)
+// - classify wrong error (real error has count=0)
+// - one caller drops (same closure, different handling at two call sites)
+// - double classify through helper (inner + outer both classify)
```
bernard-yip, line 22: Is the whole Z3/SSA harness for formally proving every error is handled exactly once needed? 

My thinking was the boundaries are what matter: skippable (continue + record), fatal (abort), and unhandled errors (no failing open) and those are already covered by the unit tests + integration tests. I don't see the need to prove every path especially at the cost of ~3k lines of SSA/Z3 code we'd have to maintain.

Thoughts?

xoverride: Good question. Let me lay out what the Z3/SSA harness actually tests, what gap it fills, and why I think the maintenance cost is lower than it looks.

**What it tests**

The property is: every AWS SDK error in every enumerator reaches ClassifySkippable exactly once on every execution path. Zero means the error can abort the pipeline (the original LAB-2525 bug). More than once means double-recording in the SkipReport and ambiguous error ownership between layers. ClassifySkippable is the single centralized funnel where every SDK error is classified, logged, and recorded — the Z3 test proves that funnel is complete and non-duplicated.

**What it provides over unit + integration tests**

The boundary behavior (skippable → continue + record, fatal → abort, unhandled → propagate) is well-covered by the existing unit and integration tests. The AST tests catch structural mistakes within a function (missing ClassifySkippable, wrong service name, missing defer Close()). Integration tests with restricted IAM roles exercise the real error paths. We could catch most issues with those alone.

Z3 adds a degree:

  - Exhaustive path coverage: Unit and integration tests exercise the paths your fixtures trigger. Z3 proves the property across all paths. You'd need a fixture for every error × service × region combination to match that coverage with integration tests.
  - Actionable diagnostics: When a violation is found, Z3 produces a counterexample — the exact execution trace showing how the property fails. It flagged the buildResource double-classify with this path:

```
  [ErrorSource] DescribeImageAttribute
    → [Wrap] fmt.Errorf
    → [Return] return err in buildResource
    → [CallResult] call result in listImagesInRegion
    → [Classify] ClassifySkippable at ec2_image_enumerator.go:134  ← 1st
    → [Return] return err in listImagesInRegion
    → [CallResult] call result in List
    → [Call] handleListError
    → [Classify] ClassifySkippable at enumerator.go:132            ← 2nd
```

  That trace told us exactly which layers were involved and can led directly to a code fix. 

AST test works on bare function names within a single file's parse tree, so it can not find this. The SSA fallback (without Z3) also missed it; its heuristic MAX logic treated the two classify branches as mutually exclusive and reported a count of 1.

  - Zero cost for new enumerators: The test auto-discovers every aws-sdk-go-v2/service/ call from the package SSA. When someone adds a new enumerator, the test verifies it automatically — no new test cases, no new fixtures.

  **Maintenance cost**

  The ~3k lines break down to: Z3 test file is ~1,600 lines, of which roughly 900 are the graph extraction and solver integration, ~450 lines of stress test patterns, and the rest ~1000 is the SSA fallback + helpers. But the key property is that the Z3/SSA test is auto-discovering — it loads the package SSA, finds all AWS SDK call sites automatically, and traces error flow from the SSA graph. It has no hardcoded enumerator names, no fixture lists, no expected-output files. When someone adds a new enumerator or changes an error path, the test auto-adapts because it reads the SSA of whatever code is compiled.

  Compare that to unit tests: every new enumerator needs new test cases for the skippable/fatal/propagation paths. Integration tests need Terraform fixtures and IAM roles. The Z3 test needs nothing — it just verifies the new code the same way it verifies existing code.

  The stress test patterns (z3_stress_enumerator_test.go, 450 lines) are the only part that might grow, and only if we want to test new adversarial patterns against the Z3 analysis itself — not per new enumerator.

  So the initial cost is high but the ongoing cost is low after the model stabilizes. The alternative — writing per-enumerator unit tests for every error path permutation — has lower initial cost but linear ongoing cost.

  Happy to discuss further.

bernard-yip: Fair enough, appreciate the writeup. I'll lay out where I land and then leave it to your call:

- Zooming out, this still feels overkill for what it protects. The whole exactly-once property is really about not double-recording in the SkipReport, and I don't think a duplicated skip entry is a problem worth a theorem prover. Its blast radius is a cosmetic dupe in a report.
- If the goal is being able to see what happened when we hit these errors in the wild, the higher-leverage fix is probably just richer debug logging on the error trace. That's what we'd actually reach for during an incident.
- I'm also not confident other engineers can maintain this as Aurelian's feature set evolves. Debugging SSA/VTA/SMT output (or an x/tools bump that shifts SSA shape) is a steep ask, and the existing knownLimitations map plus the Z3/SSA disagreement suggest the model isn't fully settled yet.
- All that said, if you feel strongly about keeping it, I won't block on it.

bernard-yip (approved):
## ✅ Approve — verified end-to-end against AWS (acct 196766918487, 3 regions)

Checked out at PR head `c2c2680` in a worktree and ran the full integration suite with real AWS credentials (`us-east-1` / `us-east-2` / `us-west-2`), deploying the restructured Terraform fixture (4 restricted IAM roles + EC2/S3/Lambda/Amplify/SSM) via a clean redeploy.

### Build & static checks
- `go build` clean; `go vet -tags=integration` clean.
- Unit suite + all **7 AST enforcement checks** pass (`TestNewEnumeratorRequiresClose`, `TestSkipReportFieldRequiresUsage`, `TestClassifySkippableServiceNames`, `TestNewEnumeratorSharesSkipReport`, `TestClassifySkippableAlwaysInIfInit`, `TestFatalErrorCodes_AllBlocked`, `TestEnumerateByARN_DispatcherExtractsServiceAndRegion`).

### Integration e2e — all resilience scenarios green
All 11 `TestSkipResilience_*` scenarios pass (`RestrictedRole`, `ByARN_Denied` / `CloudControl` / `MixedAllowDeny`, `MultiRegion`, `PerRegionDenial`, `MosaicPipeline`, `PartialEC2Access`, `WiringCheck`, `ResourcePolicyDeny`, `CloseWritesDetailFile`). Observed the intended behavior in logs, e.g. `skipped 6 operations across 3 regions` (amplify + ssm AccessDenied) with enumeration **continuing instead of aborting**. `WiringCheck` confirms every native enumerator's inner `ClassifySkippable` emits the correct short service name (no fall-through to the dispatcher safety net).

### Modules touched by this PR — all green
`TestAWSList`, `TestAWSGraph`, `TestAWSPublicResources`, `TestAWSResourcePolicies` all PASS. The `find-secrets` module runs correctly.

### Not blocking (unrelated to this PR)
- `TestAWSFindSecrets/detects_secret_in_ECS_Task_Def` — known pre-existing failure on `main` (secret-detection window; fix in flight). All other find-secrets subtests pass.
- `TestAWSIAMQuickAnalyze` — environmental (auths via the `default` profile / EC2 IMDS, unavailable in this run); `iam_quick_analyze` is not modified by this PR.

LGTM. 🚀

PR #228, opened 2026-06-11
fix(aws/iam): map CreateRole/CreateUser/CreateGroup to correct resource types

xoverride:
## Summary

- `iam:CreateRole` was mapped to resource type `service` (pattern `arn:aws:iam:*:*:*`) which never matches a role ARN. This caused `IsValidActionForResource` to reject the request before policy evaluation, producing a **false DENY** for admin principals.
- Same bug for `iam:CreateUser` and `iam:CreateGroup` — both mapped to `service` instead of their created resource type.
- Fix: `CreateRole` → `role`, `CreateUser` → `user`, `CreateGroup` → `group`.

**Impact:** False negative on CreateRole/CreateUser/CreateGroup — the evaluator (and graph module privesc detection) misses create-principal-based escalation edges. An admin with `Allow *:*` would be incorrectly denied these actions.

**Found by:** Cross-validation harness with live AWS `SimulatePrincipalPolicy` arbiter on `hengrui-sandbox-read` (352505431836). See `docs/specs/2026-06-11-aurelian-createrole-resource-type-finding.md` in `aws-iam-pe`.

## Test plan

- [x] Added regression tests for `IsValidActionForResource` with CreateRole/CreateUser/CreateGroup on their respective resource ARNs
- [x] All 2190 existing tests pass
- [x] golangci-lint clean


xoverride:
@ai-review

bernard-yip (approved):

PR #229, opened 2026-06-12
refactor(aws/privesc): drop method_NN numbering from enrichment queries (Phase 1)

bernard-yip:
## Phase 1 — Drop `method_NN` numbering from privesc enrichment queries

**Behavior-preserving rename.** No query logic changes.

### What
- Rename all **89** privesc enrichment query YAMLs: `method_NN_<slug>.yaml` → `<slug>.yaml` (via `git mv`, history preserved).
- Rewrite each `id:` from `aws/enrich/privesc/method_NN` → `aws/enrich/privesc/<slug>`.
- Collapse the per-method `order:` (was `101`–`189`) to a single band value `100`.
- Remap ~280 references across the query-loader tests, the privesc integration/pathfinding tests, the `types` examples, and the enrich README.

### Why
Method numbers collided on every new-escalation-path PR, forcing constant renumbering/rebasing. Privesc methods are **mutually order-independent** — each is an idempotent `MERGE` and no method consumes another's `CAN_PRIVESC` edge — so they only need to run *after* the `set_admin`/trust enrichers (order `0`–`13`). A shared `order: 100` band satisfies that, and new paths can now be added with just a descriptively-named file, no number.

### Behavior-preserving guarantees
- All 89 `cypher:` bodies are **byte-identical** to the previous revision (only `id:` and `order:` lines changed) — verified by both reviewers.
- Zero `method_[0-9]` tokens remain anywhere (go/yaml/md).
- The reference multiset across all test files maps 1:1 onto the 89 renamed file ids.

### Test plan
- `go build ./...` ✅
- `go test ./...` (unit) ✅
- `go vet -tags=integration ./pkg/graph/queries/... ./pkg/modules/aws/recon/...` ✅
- Neo4j privesc enrichment integration suite (`pkg/graph/queries`, testcontainers) — **169 subtests pass, 0 fail**; all 89 renamed queries load/execute; zero rename-induced failures.

### Scope note
First of a stacked series hardening the AWS privesc detector:
1. **Phase 1 (this PR)** — remove method numbers.
2. **Phase 2** — fix query correctness: cartesian-product fan-out, missing trust-policy guards, admin-as-source exclusion, lateral-vs-admin tiering, plus query-side enrichers.
3. **Phase 3** — TP + false-positive e2e fixtures (incl. trust-mismatch matrix) and full pass-rate, with a `--full` tier for expensive services.

The full Phase-0 AWS-deploying e2e baseline (which exercises the cartesian/trust FP bugs) is captured at Phase 2 kickoff.


bernard-yip:
Thanks @coderabbitai — dispositions below. This PR is a deliberately behavior-preserving rename and the base of a stacked series, so all three are being addressed in the stacked Phase-2 PR (`privesc/query-correctness`) to avoid rebasing this base.

**1. `order: 100` nondeterminism (cognito / imagebuilder / sagemaker comments).** Correct that ties sort arbitrarily, but it does **not** change results here: the privesc enrichment methods are mutually order-independent — each is an idempotent `MERGE` and none consumes another method's `CAN_PRIVESC` edge (multi-hop is computed later at analysis time). They only need to run after the `set_admin`/trust enrichers (order 0–13), which `order: 100` preserves. That said, deterministic execution is good hygiene, so we're adding a `Metadata.ID` secondary sort key in `pkg/graph/queries/queries.go` in the Phase-2 PR.

**2. FP `r.method` format mismatch (`privesc_pathfinding_test.go` 458–463 / 487–492).** Good catch — and it's a *pre-existing* weakness this rename surfaces: `r.method CONTAINS <slug>` never matched the human-readable edge property (`iam:PassRole + apprunner:CreateService`) before the rename either, so the FP branch was already vacuous. Phase 2 reworks these FP assertions to be sound — asserting on `CAN_PRIVESC` out-degree / distinct-target count (so they actually catch the cartesian fan-out) plus a stable query-id rather than the human-readable method string.

**3. Loader-test metadata assertions (`queries_test.go` 271–278).** Agreed; adding `q.Metadata.ID == id` and `q.Metadata.Order == 100` to cover the manifest contract this PR changes, in the Phase-2 PR.

Net: nothing here blocks this rename PR (behavior-preserving, CI green); the substantive query-correctness and FP-assertion work — including these three — lands in the stacked Phase-2 PR.

Elgin-Lee (approved):

PR #230, opened 2026-06-12
fix(azure/recon): per-template risk names + severity recalibration for configuration-scan

Elgin-Lee:
## Summary

Two related fixes to the Azure configuration-scan module's risk output.

### 1. Per-template risk names (was: one generic slug)

`configScanToRisk` hardcoded `Name: "azure-configuration-scan"` for every finding, collapsing distinct vulnerability classes (key-vault privilege escalation, DB firewall exposure, kusto wildcard tenants, overprivileged roles, etc.) into a single risk type. This loses signal downstream: findings can't be distinguished by name, group/dedupe together, and can't carry per-class definitions.

Now derives the name from `result.TemplateID`:
```go
Name: "azure-" + strings.ReplaceAll(result.TemplateID, "_", "-"),
```
So each template produces a distinct risk identity, e.g. `azure-key-vault-access-policy-privilege-escalation`, `azure-databases-allow-azure-services`.

### 2. Severity recalibration: High → Medium

The 8 config-scan templates that declared `severity: High` are misconfigurations / hardening gaps (DB "allow Azure services", Key Vault access-policy privesc, overprivileged custom roles, anonymous function triggers, app-service remote debugging, SSH password auth, AKS RBAC disabled / local accounts) rather than directly-exploitable High findings. Downgraded all 8 to Medium so the triage rating reflects a config-review finding. Severity flows straight from the template YAML → emitted risk, so this is a YAML-only change (+ matching integration-test assertions).

Affected templates: `aks_rbac_disabled`, `app_service_remote_debugging`, `databases_allow_azure_services`, `function_app_http_anonymous_access`, `key_vault_access_policy_privilege_escalation`, `kusto_wildcard_trusted_tenants`, `overprivileged_custom_roles`, `vm_ssh_password_authentication`.

## Scope

Intentionally **configuration-scan only**. `public-resources` uses a single-slug pattern too but that's defensible (every finding means "this resource is public", differentiated by resource); config-scan templates are distinct vulnerability classes.

## Compatibility note

The per-template naming changes the risk `Name` (identity) for configuration-scan findings. Downstream consumers keyed on the literal `"azure-configuration-scan"` will see new slugs and may need migration. The single-slug collapse was the bug.

## Tests

- Unit `TestConfigScanToRisk` asserts the derived name (`azure-test-misconfig`).
- Integration assertions derive the expected name from each block's `rc.Template`, and severity assertions updated to Medium for the 8 downgraded templates.
- `go build` / `go vet` (incl. `-tags integration`) / unit tests pass.

## End-to-end validation

Deployed to a dev Guard stack and ran `aurelian-azure-configuration-scan` against a live Azure subscription with seeded misconfigurations. Findings landed in Guard as **distinct per-template risk records** (`azure-databases-allow-azure-services`, `azure-key-vault-access-policy-privilege-escalation`, `azure-overprivileged-custom-roles`, `azure-nsg-unrestricted-port-ranges`, `azure-aks-local-accounts-enabled`, etc.) with correct titles, replacing the single generic risk. (Severity downgrade verified at the template/unit level; the live run predated that commit.)

### DIFF
--- a/pkg/modules/azure/recon/configuration_scan_integration_test.go
+++ b/pkg/modules/azure/recon/configuration_scan_integration_test.go
@@ -86,7 +86,7 @@ func TestAzureConfigurationScan(t *testing.T) {
 	// =====================================================================
 	t.Run("aks_local_accounts_enabled", func(t *testing.T) {
 		rc := findRisk(t, "aks_local_accounts_enabled", "aks_id")
-		assert.Equal(t, "azure-"+strings.ReplaceAll(rc.Template, "_", "-"), rc.Risk.Name)
+		assert.Equal(t, "azure-aks-local-accounts-enabled", rc.Risk.Name)
 		assert.Equal(t, output.RiskSeverity("low"), rc.Risk.Severity)
 		assert.Equal(t, "aks_local_accounts_enabled", rc.Template)
 		assert.NotEmpty(t, rc.Risk.Context)
@@ -97,7 +97,7 @@ func TestAzureConfigurationScan(t *testing.T) {
 	// =====================================================================
 	t.Run("app_service_auth_disabled", func(t *testing.T) {
 		rc := findRisk(t, "app_service_auth_disabled", "webapp_no_auth_id")
-		assert.Equal(t, "azure-"+strings.ReplaceAll(rc.Template, "_", "-"), rc.Risk.Name)
+		assert.Equal(t, "azure-app-service-auth-disabled", rc.Risk.Name)
 		assert.Equal(t, output.RiskSeverity("medium"), rc.Risk.Severity)
 		assert.Equal(t, "app_service_auth_disabled", rc.Template)
 
@@ -111,7 +111,7 @@ func TestAzureConfigurationScan(t *testing.T) {
 	// =====================================================================
 	t.Run("app_service_remote_debugging_enabled", func(t *testing.T) {
 		rc := findRisk(t, "app_service_remote_debugging_enabled", "webapp_debug_id")
-		assert.Equal(t, "azure-"+strings.ReplaceAll(rc.Template, "_", "-"), rc.Risk.Name)
+		assert.Equal(t, "azure-app-service-remote-debugging-enabled", rc.Risk.Name)
 		assert.Equal(t, output.RiskSeverity("medium"), rc.Risk.Severity)
 		assert.Equal(t, "app_service_remote_debugging_enabled", rc.Template)
 
@@ -129,7 +129,7 @@ func TestAzureConfigurationScan(t *testing.T) {
 	// =====================================================================
 	t.Run("databases_allow_azure_services", func(t *testing.T) {
 		rc := findRisk(t, "databases_allow_azure_services", "sql_server_id")
-		assert.Equal(t, "azure-"+strings.ReplaceAll(rc.Template, "_", "-"), rc.Risk.Name)
+		assert.Equal(t, "azure-databases-allow-azure-services", rc.Risk.Name)
 		assert.Equal(t, output.RiskSeverity("medium"), rc.Risk.Severity)
 		assert.Equal(t, "databases_allow_azure_services", rc.Template)
 	})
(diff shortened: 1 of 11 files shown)


pkg/modules/azure/recon/configuration_scan_integration_test.go
```diff
@@ -86,7 +86,7 @@ func TestAzureConfigurationScan(t *testing.T) {
 	// =====================================================================
 	t.Run("aks_local_accounts_enabled", func(t *testing.T) {
 		rc := findRisk(t, "aks_local_accounts_enabled", "aks_id")
-		assert.Equal(t, "azure-configuration-scan", rc.Risk.Name)
+		assert.Equal(t, "azure-"+strings.ReplaceAll(rc.Template, "_", "-"), rc.Risk.Name)
```
bernard-yip: These updated assertions (this one and the other 12 below) re-derive the expected value with the same `"azure-"+strings.ReplaceAll(...)` transformation the production code uses, so a bug in the transformation can never fail the test. Hardcoded literals are stronger and self-documenting:

```go
assert.Equal(t, "azure-aks-local-accounts-enabled", rc.Risk.Name)
```

bernard-yip (commented):
Two findings, both on the integration test — see the inline comments: a blocking stale invariant at line 223 that still asserts the old `azure-configuration-scan` name, and the updated name assertions being tautological (they re-derive the expected value with the same transformation as production). Everything else looks good.

pkg/modules/azure/recon/configuration_scan_integration_test.go
```diff
@@ -211,8 +211,8 @@ func TestAzureConfigurationScan(t *testing.T) {
 	// =====================================================================
 	t.Run("vm_ssh_password_authentication", func(t *testing.T) {
 		rc := findRisk(t, "vm_ssh_password_authentication", "vm_id")
-		assert.Equal(t, "azure-configuration-scan", rc.Risk.Name)
-		assert.Equal(t, output.RiskSeverity("high"), rc.Risk.Severity)
+		assert.Equal(t, "azure-"+strings.ReplaceAll(rc.Template, "_", "-"), rc.Risk.Name)
+		assert.Equal(t, output.RiskSeverity("medium"), rc.Risk.Severity)
 		assert.Equal(t, "vm_ssh_password_authentication", rc.Template)
```
bernard-yip, line 216: Blocking — just below this hunk, at line 223, the cross-finding invariant was missed by this PR and still asserts the old name for every risk:

```go
t.Run("all risks have name azure-configuration-scan", func(t *testing.T) {
    for _, rc := range all {
        assert.Equal(t, "azure-configuration-scan", rc.Risk.Name, ...)
    }
})
```

The integration suite is guaranteed to fail once it actually runs — `go vet -tags integration` compiles the test code but doesn't execute it, which is why this slipped through. Update the invariant to the derived-name form or drop the subtest.

Elgin-Lee: Good catch — fixed in 19a67d9d. Updated the invariant to validate the per-template naming (`azure-`+derived) instead of the stale literal. Root cause: my slug-change find/replace matched the bare `assert.Equal(t, "azure-configuration-scan", rc.Risk.Name)` form but this block uses the extra-args signature, so it didn't match — and `go vet -tags integration` compiles without executing, so the contradiction never surfaced. You're right that compile ≠ run.

pkg/modules/azure/recon/configuration_scan_integration_test.go
Elgin-Lee: Agreed — fixed in 19a67d9d. Replaced all 12 per-template name assertions with hardcoded literals (e.g. `assert.Equal(t, "azure-aks-local-accounts-enabled", rc.Risk.Name)`) so a bug in the production naming transform actually fails the test. Left the cross-finding invariant loop deriving the expected value since it can't hardcode per-finding.

pkg/modules/azure/recon/configuration_scan_integration_test.go
Elgin-Lee: Follow-up: rather than re-derive the name in the invariant (which would've been tautological the same way the per-case ones were), swapped it for a negative regression guard — `assert.NotEqual(t, "azure-configuration-scan", rc.Risk.Name)` over all findings. Guards the exact regression this PR prevents (collapse back to one slug) across every finding, and the production transform can't self-satisfy it. Force-pushed as 7e1bea01.

Elgin-Lee: Final: removed the invariant subtest entirely (your 'or drop it' option). The 12 per-template assertions now use hardcoded literals, which is the real coverage; a derived/negative invariant added little over that. Force-pushed as b30a4ed6.

bernard-yip (approved):

PR #232, opened 2026-06-12
feat(aws/privesc): query correctness — cartesian fix + structural guards (Phase 2)

bernard-yip:
## Phase 2 — AWS privesc query correctness (cartesian fix + structural guards)

Targets `main`. Phase 1 (#229, the `method_NN` rename) is already merged; this Phase-2 work is rebased on top of it.

### What this does
Fixes the privilege-escalation enrichment queries so a `CAN_PRIVESC` edge means a *real, scoped* escalation instead of an all-principals fan-out, across all 89 methods.

**Enrichers (guard inputs):**
- Fix `set_admin` — it ran `ANY()` over a JSON-*string* and never matched, so `_is_admin` was effectively dead on real data (the root cause of admins fanning out).
- Extend the role-trust enricher to emit `CAN_ASSUME` for account-root trust (`:root`), not just explicit-principal.
- Reliable instance→role `HAS_ROLE`; new `resource_service_role.yaml` enricher (CFN/CodeBuild/Glue/AppRunner/ECS/StepFunctions/SageMaker/DataPipeline → service role).

**Methods (all 89):**
- Replace the `WITH attacker → MATCH (victim:Principal) → MERGE`-to-everyone fan-out with a scoped edge to the correct target (self-loop / passed role / `CAN_ASSUME`-reached role / resource's `HAS_ROLE` role / stub).
- Cross-cutting guards: admin-as-source exclusion, self-target exclusion, lateral-vs-admin severity tiering (admin→high / privileged→medium / else→low; suppress non-privileged except trust-backed takeover).
- Per-technique guards from the FP-guard catalog: service-trust match for PassRole methods, instance-profile for EC2/EMR/ImageBuilder/GameLift, customer-managed+attached for policy-version, USER-scoping for access-key/login-profile, `CAN_ASSUME` requirement for AssumeRole-style.

**CodeRabbit (#229) fixes folded in:** deterministic `Metadata.ID` sort tiebreak; loader-test metadata assertions; reworked the vacuous FP assertion (matches the exact `pe.method` literal).

### Results (measured on a real account, `bfr`)
| Metric | Before | After |
|---|---|---|
| Total `CAN_PRIVESC` edges | 9,628 | **441** (−95.4%) |
| Sources reaching ≥75 of 150 principals | 56 | **1** |
| False-positive principals firing | 5 | **0 / 18** |

### Validation
- **Neo4j synthetic tests** (real `EnrichAWS` pipeline, shared all-guards-satisfied seed): all 89 methods fire exactly one correctly-scoped edge with a per-method no-fan-out assertion; FP/negative cases tightened so the named guard is the sole rejection. All 11 groups green.
- **AWS e2e** (`bfr`): cartesian collapse + 0 false positives confirmed on real data.
- **IAM policy evaluator (`pkg/aws/iam/evaluator.go`) untouched.**

### Scope note — true-positive detection is validated in Phase 3
The existing `privesc-pathfinding` terraform fixture predates these guards, so it can't exercise the positive path: only 10/87 TP cases fire, and the **77 "misses" are fixture deficiencies** (the fixture's roles lack service-trust / `CAN_ASSUME` / privileged-target / paired-permission structure that real exploitable paths have) — **not query bugs** (each was root-caused; 0 false positives, 0 category-A query bugs). Phase 3 rebuilds the fixture (trust matrix + TP/FP per technique) for full real-data TP+FP validation; those 77 are its itemized work list. This PR's TP confidence rests on the synthetic tests (which prove each method fires given guard-satisfying inputs) + the demonstrated fan-out collapse and FP suppression on real data.


bernard-yip:
**Re: the LOW finding on the unreachable `ELSE 'low'` in the compute-takeover severity `CASE`** — leaving this as-is, by design.

The observation is correct: in the compute-takeover methods the `WHERE` requires `_is_admin OR _is_privileged`, so the `ELSE 'low'` branch is unreachable. But the same three-way severity `CASE` is a shared template across all 26 method files that carry it — including the trust-backed direct-takeover (AssumeRole-style) methods, where the `low` tier **is** reachable by design (our "fail-open at `low` for direct-takeover" decision). Collapsing just the compute-takeover files to a two-way `CASE` would diverge them from the shared template for a purely cosmetic gain, so we're keeping the uniform `CASE` for maintainability.

The other four findings (1 `set_admin` prefix-collision, 2 `UpdateLoginProfile` CreateLoginProfile precondition, 3 lambda asymmetric same-account coalesce, 5 EC2 cross-product ordering) are valid and addressed in a follow-up commit.

bernard-yip:
Addressed both review rounds in `2870e206`, `48f138ed`, `887711ee`.

**AI Review (github-actions):**
- `set_admin` — anchored to the AWS-managed policy ARN so scoped/look-alike policies aren't flagged `_is_admin`.
- `iam_update_login_profile` — dropped the `IAM_CREATELOGINPROFILE` precondition (it was a false negative on the common path; profile-existence stays an out-of-graph deferred item).
- lambda methods (×5) — symmetric `coalesce(attacker.Arn, attacker.arn)` in the same-account compare.
- EC2 compute-takeover (×2) — reordered to attacker-first + `WITH DISTINCT` to drop the cross-product.
- The unreachable `ELSE 'low'` is intentional template uniformity (explained [above](#issuecomment-4693393822)).

**CodeRabbit:**
- Quote-bound the prefix-collision substring matches in `extract_role_trust_relationships`, `resource_service_role`, `resource_to_role`, and `set_admin`.
- `codestar_create_project` — constrained target to `:Principal`.
- `iam_add_user_to_group` — scoped to the permission-target group + deterministic `max()` severity.
- `cloudformation_update_stack` — verified the `UPDATESTACK` edge resolves to a service stub for `Resource:*` policies, so I added a **same-account guard** rather than carrying the stack through (the suggested stack-carry would cause false negatives).
- Consistency/doc: `cloudformation_update_stackset` name, `emr_run_job_flow` SLR exclusion, `batch_submit_job`/`bedrock_access_code_interpreter` account constraints, README inventory (28 → 89), and a same-`Order` `Metadata.ID` tiebreaker regression test.

**Deferred to Phase 3:** the FP-test-soundness item on `privesc_pathfinding_test.go` (method-string matching on deduped `CAN_PRIVESC` edges) — it overlaps the upcoming real-data fixture rebuild and is tracked there.

Full synthetic Neo4j suite is green (89 per-method subtests + the analysis-query e2e); loader/tiebreaker units pass; `go test ./...` clean.

bernard-yip:
## Added: multi-edge `CAN_PRIVESC` (one edge per method) — `9f2541a1`

Beyond the review-fix commits above, this PR now also changes the `CAN_PRIVESC` edge model so **all** privilege-escalation method-paths between two principals can be enumerated, not just the one that happened to win.

**Before:** every method did `MERGE (a)-[pe:CAN_PRIVESC]->(t) ... SET pe.method='<M>'`. When several methods reached the same `(attacker, target)` pair they collapsed onto one edge with a last-write-wins `pe.method` — so the graph only ever showed one technique per pair.

**After:** `method` is part of the edge identity — `MERGE (a)-[pe:CAN_PRIVESC {method:'<M>'}]->(t)` — so each distinct technique is its own edge. Severity stays per-edge.

Changes:
- **88 method YAMLs** rewritten to the `{method:'…'}` MERGE form (method strings byte-identical to before; verified). `iam_create_service_linked_role` stays an explicit no-op.
- **`analysis/aws/privesc_paths.yaml`** now returns the per-hop method sequence (`[r IN relationships(path) | r.method] AS methods`), dedups by distinct `(attacker, target, methods)`, with a deterministic `ORDER BY` + `LIMIT 2000` path-explosion guardrail.
- **Tests:** per-method validation matches edges *by* method; recon edge-count assertions moved to `count(DISTINCT target)` (the correct "how many principals" measure under multi-edge); the FP-check regex was updated to read the new `{method:'…'}` form. This makes the per-method FP assertion **sound** — previously a forbidden method could be masked by edge dedup (the deferred CodeRabbit concern on `privesc_pathfinding_test.go`). New `TestPrivescMultiEdgePerMethod` proves 2 methods → 2 edges → 2 analysis method-paths.

**Validation:** full synthetic Neo4j suite green (205 subtests, 0 failures); both reviewers approved.

**Known follow-up (not in this PR):** `analysis/aws/privesc_paths.yaml` has no production Go consumer today (`AnalyzeAWS` is currently unwired; the formatter only runs `EnrichAWS`), so the new `methods` column is queryable in the graph but not yet surfaced in module output. Wiring it into a finding/proof is net-new scope, tracked for the Phase-3 work.

bernard-yip:
Addressed the latest review findings in `da67a181`.

- **`resource_to_role.yaml` (Major — name-form `HAS_ROLE` false negative):** correct catch. The previous quote-binding (`CONTAINS '"' + IamInstanceProfile + '"'`) only matched the NAME form if the role's `InstanceProfileList` serialized `InstanceProfileName`, which isn't guaranteed — and CloudControl EC2 instances commonly carry the profile **name**. Added a second, closing-quote-anchored clause matching the name against the ARN's trailing segment:
  ```
  role.InstanceProfileList CONTAINS ('"' + resource.IamInstanceProfile + '"')
  OR role.InstanceProfileList CONTAINS ('instance-profile/' + resource.IamInstanceProfile + '"')
  ```
  The trailing `"` keeps prefix-collision protection (`instance-profile/p` won't match `instance-profile/p-2"`). New `TestResourceToRoleNameFormHasRole` seeds the NAME form against an ARN-only `InstanceProfileList` (worst case) and asserts `HAS_ROLE` — it fails on the old single-clause query and passes with the fix, so the previously-untested branch now has coverage.
- **`README.md` (Minor):** dropped the stale "DataPipeline" from the service-bucket heading (`iam_pass_role_datapipeline` lives in the IAM section); inventory still sums to 89.
- **`queries_test.go` (Minor):** the tiebreaker test now snapshots and restores prior registry entries instead of unconditionally deleting the injected IDs.

**Deferring the `privesc_integration_test.go:330` fan-out-allowlist suggestion** to the fixture-rebuild work: that assertion only runs against the live AWS `privesc-pathfinding` Terraform fixture, and the "known passable-role allowlist" it wants is defined by that fixture — which is being rebuilt next. We'll tighten `count(DISTINCT target) <= 10` into an exact expected-target-set assertion as part of that rebuild, when the allowlist is known.

Full synthetic Neo4j suite green (206 cases, 0 failures); both reviewers approved.

Elgin-Lee (approved):
Approving to land the privesc Phase-2 query-correctness foundation. Behavior validated transitively via the stacked #233 branch (15-account live run); follow-up precision/recall items tracked under LAB-3742 (LAB-4176–4184).

PR #233, opened 2026-06-12
feat(aws/privesc): close detection gaps, add resource collectors + precision guards, surface privesc-path findings

bernard-yip:
## Summary

Hardens Aurelian's AWS privilege-escalation detection (Neo4j `CAN_PRIVESC` graph enrichment) and surfaces the results as findings. Stacked on the cartesian-fix / structural-guard work in the base branch.

### Detection coverage
- Closes gaps where a privesc method's guard required an action edge or resource the engine never produced: adds the missing IAM / CloudFormation / DataPipeline / Bedrock-AgentCore / Cognito-Identity actions to the evaluator's action-coverage tables (allowlist + action→resource map), collects EC2 instances, and surfaces role `Principal.Federated` trust (used by Cognito identity-pool roles).
- Adds recon resource collectors so existing-compute takeover paths fire on real accounts (previously they only fired against hand-seeded test data): CloudFormation stacks & stack sets, Batch job definitions, Bedrock AgentCore code interpreters, CodeBuild projects, Glue jobs & dev endpoints, App Runner services, ECS task definitions, Step Functions state machines, SageMaker notebook instances. Each emits the backing resource's service role so the `resource_service_role` / `resource_to_role` enrichers build the `HAS_ROLE` edge the takeover methods traverse.
- Fixes the `gamelift_create_fleet` guard: it required an EC2 instance profile on the passed role, but a GameLift fleet's instance role is assumed via `gamelift.amazonaws.com` trust and carries no instance profile — the guard suppressed the real escalation. Now aligned with its `gamelift_createbuild_createfleet` sibling (trust discriminator + service-linked-role exclusion + same-account check).

### Precision
- Collects per-user login-profile existence and active access-key count; tri-state fail-open guards on `iam_update_login_profile` and `iam_create_access_key`.
- Surfaces managed-policy version count; guards `iam_create_policy_version` (suppressed at the 5-version limit).
- Re-points the change-set / Batch / Bedrock takeover methods from a service-stub terminal at the specific backing resource's role.

### Output
- New `aurelian aws analyze graph` module: runs the privesc-path analysis over the enriched graph and emits one finding (`output.AurelianRisk`) per attacker→target path (method sequence + hop count + severity).
- `aurelian aws recon graph --neo4j-uri …` now seeds Neo4j **additively** — it writes the normal JSON output **and** populates the graph — so the two commands compose (`recon graph` to populate, `analyze graph` to report).

### Tests
- Rebuilt the `privesc-pathfinding` terraform fixture with a per-technique true-positive role for each method plus a false-positive matrix (missing-permission, trust-policy-mismatch, non-privileged-target).
- Live end-to-end run on the integration account; a synthetic Neo4j seeded suite validates the enrichment Cypher without AWS, replayable from a captured recon snapshot.
- **The `--full` heavyweight tier is now live-validated** (previously skip-logged behind `AURELIAN_E2E_FULL=1`): EMR, EMR-Serverless, GameLift, ImageBuilder, Braket, Omics, KinesisAnalytics plus the concrete-ARN App Runner / SageMaker cases now provision (via `full.tf`, gated on `var.enable_full`) and pass. Running the full tier for the first time fixed 6 pre-existing full-tier bugs:
  - `compute_admin` now trusts `sagemaker.amazonaws.com` so the real SageMaker notebook (which runs it) can be created;
  - `fulladmin-emr` / `fulladmin-imagebuilder` get ec2 trust + instance profiles (EC2-backed escalation targets, as the queries require);
  - the `gamelift_create_fleet` guard fix above;
  - the synthetic/real App Runner node collision in the snapshot/replay harness (`SyntheticActionEdges` now lets a real collected resource override only types with no synthetic stand-in);
  - `directTrustAdminRoleARN` is now serialized into the recon snapshot, so the AWS-free replay can resolve the F6 `sts_assume_direct_trust` target.
- **Result: live `--full` e2e 160 PASS / 0 FAIL / 0 SKIP; AWS-free replay + freshness 140 / 0 / 0.** The recon snapshot was regenerated from a real `--full` capture.

The IAM policy-evaluation algorithm is unchanged; coverage is extended only via additive data tables, new collectors, a transformer signal, and Cypher.


Elgin-Lee (approved):
Approving to land the privesc collectors + analyze-graph module on top of the merged #232 foundation. Validated via a 15-account live run; precision/recall follow-ups tracked under LAB-3742 (LAB-4176–4184).

PR #235, opened 2026-06-15
fix(aws/iam): remove IsValidActionForResource pre-filter from Evaluate

xoverride:
## Summary

- Remove `IsValidActionForResource` gate from `PolicyEvaluator.Evaluate()` — it incorrectly denied actions on concrete resource ARNs when the action's resource-type map didn't include a matching pattern, even when the principal's policy grants `Resource:"*"`.
- This caused **false negatives** for `lambda:CreateFunction`, `sts:AssumeRole`, `ec2:RunInstances`, and any action whose resource-type map only included `"service"`.
- The statement-level evaluation already handles resource matching correctly (`Resource:"*"` matches any ARN). The pre-filter was a misguided optimization.

**Impact:** On the 137-principal sandbox, Aurelian fired only 10/44 escalation rules (297 paths) vs IAMSpy's 44 rules (2099 paths). The entire gap was this erroneous pre-filter denying legitimate permissions.

**Found by:** Three-backend cross-validation harness (policy_summary vs IAMSpy vs Aurelian) on `hengrui-sandbox-read` (352505431836). See `docs/specs/2026-06-15-AURELIAN-BUG-check-specific-service-arn.md` in `aws-iam-pe`.

**Supersedes:** #228 (`CreateRole`/`CreateUser`/`CreateGroup` resource-type map fix) — that fix addressed individual mappings; this fix removes the flawed gate entirely, covering all actions.

## Test plan

- [x] Updated test that asserted buggy behavior (`TestPolicyEvaluator_InvalidActionForResource` → `TestPolicyEvaluator_ActionOnConcreteResourceARN`)
- [x] Updated test asserting `CreateFunction` on function ARN is denied → now asserts allowed
- [x] All 2190 existing tests pass
- [x] golangci-lint clean

### DIFF
--- a/pkg/aws/iam/evaluator.go
+++ b/pkg/aws/iam/evaluator.go
@@ -181,14 +181,11 @@ func policyToStatementList(policy *types.Policy) *types.PolicyStatementList {
 
 // Evaluate performs the full policy evaluation
 func (e *PolicyEvaluator) Evaluate(req *EvaluationRequest) (*EvaluationResult, error) {
-	// First validate that the action is valid for the resource type
-	if !IsValidActionForResource(req.Action, req.Resource) {
-		return &EvaluationResult{
-			Allowed:           false,
-			PolicyResult:      NewPolicyResult(),
-			EvaluationDetails: fmt.Sprintf("Action %s is not valid for resource %s", req.Action, req.Resource),
-		}, nil
-	}
+	// NOTE: IsValidActionForResource was previously used here as a pre-filter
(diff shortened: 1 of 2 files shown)


xoverride:
Closing — the IsValidActionForResource pre-filter is redundant but harmless on main (all callers pre-match resources via GetResourcesByAction). The bug only manifests in the subprocess evaluator on feat/escalation-solver-model-export, where external callers send arbitrary (action, resource) pairs. Fix stays on the feature branch.

PR #238, opened 2026-06-24
chore(deps): add Dependabot config scoped to titus

xoverride:
## What

Adds `.github/dependabot.yml` to automatically open PRs when our one internal Praetorian dependency, `github.com/praetorian-inc/titus`, cuts a new release.

## Why

`titus` is currently pinned to a commit-based pseudo-version (`v1.1.10-0.20260324...`) and is behind the latest tagged release (`v1.2.5`). It's the only `praetorian-inc/*` module in our dependency graph (imported across ~35 files under `pkg/secrets/`). A scoped Dependabot config keeps it current with minimal maintenance.

## Scope

- `gomod` ecosystem, weekly schedule
- `allow` list restricts updates to **titus only** — the large AWS/Azure SDK surface is intentionally excluded to avoid PR noise
- Bump PRs are validated by existing CI (`go-ci.yml`)

## Verification

- YAML validated locally (parses, correct schema)
- Dependabot activates its scan loop once this lands on `main`; the first titus bump PR (likely → v1.2.5) confirms it works end to end

PR #241, opened 2026-06-25
fix(find-secrets): stream JSONL output to bound memory

xoverride:
## Summary

Streams module results to disk as JSONL instead of accumulating all results in
memory via `Collect()` + `json.MarshalIndent`. This bounds find-secrets peak RSS
to a small constant regardless of finding count, and additionally fixes three
error-handling bugs. Neo4j seeding (`--neo4j-uri`) is preserved unchanged.

**This PR supersedes the first half of #156** (the streaming/output memory fix).
The second half of #156 (the titus bump) has already landed via **#238** and
**#239** (titus is now v1.2.6). #156 can therefore be closed.

## What changed

- **New `Sink` abstraction** (`pkg/plugin/sink.go`): `JSONLSink` streams one JSON
  object per line, creating the file lazily on first write; `Neo4jSink` buffers
  graph entities and seeds via the existing `GraphFormatter` on close.
- **`runModule` rewritten** to fan each streamed result out to the sink list
  instead of materializing the full slice. Output extension `.json` -> `.jsonl`.
- **`risk.go`**: `json.MarshalIndent` -> `json.Marshal` for proof bytes (compact;
  consumers parse by key, so whitespace is irrelevant).

## Bugs fixed (regressions present in #156's approach)

1. **Partial file on error** -- a producer error mid-stream now removes the
   partial `.jsonl` (Abort) instead of leaving truncated output that looks valid.
2. **Goroutine leak / swallowed error** -- a write error now drains the pipeline
   and calls `Wait()`, so the unbuffered-channel producer can't block and the
   upstream error is surfaced.
3. **Empty-file churn** -- zero results create no file and no directory (lazy
   create), instead of create-then-remove.

## Memory & performance evidence

Real process peak RSS (`/usr/bin/time -l`), old `Collect()` vs new streaming,
**both on titus v1.2.6** so this isolates the output-path change:

| Findings | OLD (Collect+pretty) | NEW (streaming) | Reduction |
|----------|----------------------|-----------------|-----------|
| 100,000  | 115 MB               | 26 MB           | 4.4x      |
| 1,000,000| 1,071 MB             | 28 MB           | 38x       |

OLD grows linearly with finding count; NEW is flat (~27 MB). Go benchmark agrees
(`B/op`: 35 MB -> 5 KB at 100k) and shows **28-48% faster** wall-clock -- streaming
drops slice accumulation and indentation.

Why this is still needed after the titus bump: titus v1.2.6 fixed the matcher's
allocation *churn*; this fixes result *accumulation* (retained RSS), which titus
does not touch.

## Testing

- Unit tests: `JSONLSink` (write/lazy-create/abort), `Neo4jSink` (content gate,
  no-op on empty, seed on entities), benchmark.
- Hermetic e2e through real `runModule`: N findings -> valid `.jsonl`; zero
  results -> no file; producer error -> no partial file.
- Live AWS (sandbox): full scan detected all planted secrets across every
  resource type -> valid `.jsonl`.
- Full suite + `golangci-lint` clean; `go mod tidy` no-op.

## Viewing output (human-readable)

The canonical output stays JSONL (one object per line) so it streams with O(1)
memory and works with line-based tooling. For a human-readable view, pretty-print
on demand with `jq`:

```
jq . <module>-<timestamp>.jsonl
```

A pretty-printed *array* file is intentionally NOT the default: serializing one
big indented array is exactly the accumulation this PR removes. If a write-time
`--format pretty` is wanted later, it should stream an indented array
incrementally (or be explicitly opt-in for small scans) and ship as its own PR.

## risk.go proof change — Guard impact verified

The `MarshalIndent` -> `Marshal` change for the secret proof is whitespace-only (JSON-semantically identical). Verified against the Guard repo that consumes it:

- Guard UI (`ui/src/sections/poe/processors/Secrets.processor.tsx`) `JSON.parse`s the proof and accesses fields structurally — whitespace-agnostic.
- Guard backend (`backend/pkg/tasks/capabilities/secrets/invoke.go`) already emits its own secret proofs with compact `json.Marshal`, so compact is the established format; aurelian was the outlier with indentation.
- No hashing / byte-exact / regex-over-raw-bytes handling exists in the proof path on either side.

Conclusion: no impact on Guard — the change aligns aurelian with Guard's own serialization convention.


cmd/generator.go
```diff
@@ -286,69 +285,83 @@
...
+		if username == "" {
+			username = "neo4j"
 		}
-
-		// Ensure the output directory exists
-		if err := utils.EnsureFileDirectory(outputPath); err != nil {
-			return fmt.Errorf("failed to create output directory: %w", err)
+		password, _ := argsMap["neo4j-password"].(string)
+		if password == "" {
+			password = "neo4j"
```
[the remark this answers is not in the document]
xoverride: Default. Not an issue

xoverride:
@ai-review

pkg/plugin/sink.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// Abort closes and removes the temp file, leaving the final path untouched.
+// No-op if nothing was written.
+func (s *JSONLSink) Abort() error {
+	if s.f == nil {
+		return nil
+	}
+	tmp := s.tmp
+	_ = s.w.Flush()
```
xoverride: **Minor:** `Abort()` flushes the `bufio.Writer` to a temp file it's about to `os.Remove`. The `Flush()` is pointless on the failure path — and on a full disk it would surface an error you then discard with `_ =`. Can drop the `Flush()` and just `Close()` + `Remove()`. Harmless to correctness.

pkg/plugin/sink.go
```diff
@@ -0,0 +1,182 @@
...
+	if s.f != nil {
+		return nil
+	}
+	if err := utils.EnsureFileDirectory(s.path); err != nil {
+		return fmt.Errorf("creating output directory: %w", err)
+	}
+	// Stage in a sibling temp file so the final path is only written atomically
+	// on success (see Close); a mid-stream failure cannot truncate or delete an
+	// existing file at s.path.
+	f, err := os.CreateTemp(filepath.Dir(s.path), filepath.Base(s.path)+".*.tmp")
```
xoverride, line 67: **Minor (crash-only):** `os.CreateTemp` in the output dir means an ungraceful kill (SIGKILL / power loss) leaves `<module>-<ts>.jsonl.<rand>.tmp` litter next to real output. The graceful Close/Abort paths clean up, so this is only the hard-crash case. Same-dir temp is required for the atomic rename, so this is an acceptable tradeoff — worth a one-line note in the doc comment, or a startup sweep of stale `*.tmp` if it ever bothers users.

cmd/generator.go
```diff
@@ -286,85 +285,88 @@ func runModule(cmd *cobra.Command, module plugin.Module, platform plugin.Platfor
...
-				return fmt.Errorf("connecting to Neo4j: %w", err)
+	// Stream each result to every sink as it is produced.
+	count := 0
+	for item := range p2.Range() {
+		for _, s := range sinks {
+			if err := s.Write(item); err != nil {
+				// Drain the rest so the unbuffered pipeline producer never blocks,
+				// then surface both the write error and any producer error.
+				for range p2.Range() {
+				}
```
xoverride: **DRY:** this hand-rolled drain duplicates `pipeline.Drain()` (`pkg/pipeline/pipeline.go:93`), which is exactly consume-and-discard-then-`Wait()`. This block reduces to `errors.Join(fmt.Errorf("writing output: %w", err), p2.Drain())`, matching the pipeline's own vocabulary and dropping the manual loop + separate `Wait()`.

xoverride (commented):
Overall a strong PR — the streaming `Sink` fan-out + atomic temp-file-rename `JSONLSink` is sound, the memory evidence is real (process RSS, not just B/op), and test coverage is genuinely good (hermetic e2e through real `runModule`, atomic-replace, abort-preserves-existing, empty-no-file, unreachable-Neo4j). CI Lint/Test were pending (not failed) at review time.

Findings below are all minor — none block merge. The two correctness items are defensive/cosmetic, not bugs.

**One semantic change to sign off on:** Neo4j is now an additive best-effort sink — an unreachable `--neo4j-uri` logs and continues instead of hard-failing the run (previously it failed). Intentional, documented, and tested; just confirm it matches product expectations.

Worth calling out as done right: `errors.Join` on all close paths (no swallowed errors), `%w` throughout, atomic replace with the pre-existing-file invariant proven by `TestJSONLSink_AbortPreservesExistingFile`, and deleting the test-only `resultsContainGraphEntities` gate in favor of real fan-out via `isGraphEntity`.

cmd/generator_e2e_test.go
```diff
@@ -0,0 +1,163 @@
...
+	const n = 5
+	mod := &testutils.MockModule{
+		IDValue:       "e2e-stream",
+		PlatformValue: plugin.PlatformAWS,
+		CategoryValue: plugin.CategoryRecon,
+		RunFn: func(cfg plugin.Config, out *pipeline.P[model.AurelianModel]) error {
+			for i := 0; i < n; i++ {
+				out.Send(output.AurelianRisk{
+					Name:               "aws-secret-finding",
+					Severity:           output.RiskSeverityHigh,
```
xoverride: **Go 1.24+ nit:** `for i := 0; i < n; i++` → `for i := range n`. Same at line 140. Test-only, but flagging the legacy loop form per the modern-Go idioms checklist.

xoverride:
@ai-review

xoverride:
@ai-review

xoverride:
@ai-review

PR #242, opened 2026-06-26
Make GCP Accept Single Resource Targets

josephwhenry:


pkg/secrets/scanner.go
```diff
@@ -105,22 +105,29 @@ func (s *SecretScanner) Scan(input output.ScanInput, out *pipeline.P[SecretScanR
...
+// ScanFlushAndClose runs Scan over every input, flushes any timed-out retry
+// matches into the same output pipeline, closes the scanner, then closes out.
+// It is a drop-in replacement for `pipeline.Pipe(in, s.Scan, out, opts...)`
+// when the caller is done with the scanner. The drain runs only after every
 // Scan has completed, since the matcher's queue is shared across all scanned
-// blobs.
-func (s *SecretScanner) ScanAndFlush(in *pipeline.P[output.ScanInput], out *pipeline.P[SecretScanResult], opts ...*pipeline.PipeOpts) {
+// blobs. Closing the scanner before closing out ensures downstream consumers
+// do not observe completion while the Titus database is still open.
+func (s *SecretScanner) ScanFlushAndClose(in *pipeline.P[output.ScanInput], out *pipeline.P[SecretScanResult], opts ...*pipeline.PipeOpts) {
```
xoverride, line 81: Optional (nit): `ScanFlushAndClose` is now single-shot — it closes the scanner. A second call (or a call after an explicit `Close`) sets `s.ps == nil`, and the spawned `s.Scan` goroutine then nil-derefs `s.ps` → SIGSEGV. No current caller does this (all three modules call it once), so this is latent, but folding `Close` in here makes "call it twice" a more natural mistake than before. Consider failing fast at the entry:

```suggestion
func (s *SecretScanner) ScanFlushAndClose(in *pipeline.P[output.ScanInput], out *pipeline.P[SecretScanResult], opts ...*pipeline.PipeOpts) {
	if s.ps == nil {
		out.CloseWithError(fmt.Errorf("secret scanner not started or already closed (ScanFlushAndClose is single-shot)"))
		return
	}
```

This fails deterministically (even on empty input) before any goroutine spawns, and surfaces through `out.Wait()`.

pkg/secrets/scanner.go
```diff
@@ -105,22 +105,29 @@ func (s *SecretScanner) Scan(input output.ScanInput, out *pipeline.P[SecretScanR
...
 	scanned := pipeline.New[SecretScanResult]()
 	pipeline.Pipe(in, s.Scan, scanned, opts...)
```
xoverride, line 88: Optional (nit), related: `s.Scan` is exported and exercised directly in unit tests, so it's effectively a public pipeline stage. On a closed/unstarted scanner it nil-derefs `s.ps.scanContent(...)` rather than erroring — worse, it would silently drop content in a *secret* scanner. Suggest a matching guard at the top of `Scan` (before the path-filter check):

```go
if s.ps == nil {
	return fmt.Errorf("secret scanner not started or already closed")
}
```

This keeps the nil-`s.ps` contract coherent across the type: `Scan` errors, `ScanFlushAndClose` errors, and `Flush` stays a benign no-op (already guarded). `fmt` is already imported; existing `require.NoError(s.Scan(...))` tests still pass since they start the scanner first.

xoverride (dismissed):
This review the scanner part of the changes instead of the full PR.
LGTM — the `ScanAndFlush` → `ScanFlushAndClose` refactor correctly fixes the race between the flush goroutine and the module's deferred `s.Close()` by sequencing the close inside the goroutine (drain → close DB → close out). Verified the close ordering, error propagation, and all three modules reaching `ScanFlushAndClose` with no early-return leak; full suite + `-race` pass.

Approving. Two **optional, non-blocking** hardening nits inline — both guard a latent nil-`s.ps` deref that no current caller hits (all three modules call `ScanFlushAndClose` exactly once), so they're insurance against future misuse, not merge blockers.

Tanq16:
@ai-review

Tanq16 (approved):

PR #252, opened 2026-07-15
Enumerate AWS RAM resource shares via list-all

ztgrace:
## Summary

Adds native AWS Resource Access Manager (RAM) resource-share enumeration to Aurelian's `list-all` recon path.

A new `RAMResourceShareEnumerator` (`pkg/aws/enumeration/ram_resourceshare_enumerator.go`) uses the native RAM SDK — `GetResourceShares` (`ResourceOwner=SELF`) then `GetResourceShareAssociations` for principals and resources — and emits one `output.AWSResource` per share carrying `AllowExternalPrincipals`, `Status`, `FeatureSet`, `Principals`, and `ResourceArns`.

- Registered on the enumeration dispatcher, so `list-all --scan-type full` requests it.
- `AWS::RAM::ResourceShare` added to the resource-type baseline union.
- `ram` ARN-resolver entry (`resource-share/` prefix) so by-ARN lookups route to the native enumerator instead of degrading to the CloudControl fallback — CloudControl's RAM read/list handlers hold only `ram:GetResourceShares` and cannot populate the associations.

## Why native SDK, not CloudControl

CloudControl surfaces `AllowExternalPrincipals` for free but cannot return the associated principals/resources (its read handler lacks the association permissions), and consumer-side visibility needs the SDK regardless — so a single native collector is simpler than CloudControl + a separate enrichment pass.

## Test plan

- Unit tests: pure `buildResourceShareResource` builder (external / org-only / missing-owner) and `EnumerateByARN` error branches (bad ARN, non-resource-share RAM ARN, missing region).
- Integration: terraform fixture (`test/terraform/aws/recon/ram-resource-share/`, randomized share names) + `//go:build integration` test exercising `EnumerateAll` — ran live against real RAM infra.
- `go build ./...` and the `pkg/aws/enumeration`, `pkg/aws/resourcetypes`, `pkg/types` suites pass, including the resource-type coverage/format/sort gates.

## Out of scope (follow-on)

- Finding generation: flagging `AllowExternalPrincipals=true` / principals outside the org as a risk in `public-resources` / `ResourceEvaluator`.
- Consumer-side shares (`ResourceOwner=OTHER-ACCOUNTS`, accepted external shares).

## Linear

Implements LAB-4944 (child of LAB-780 — AWS Attack Capabilities / Aurelian).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,7 +36,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/synapse/armsynapse v0.8.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/trafficmanager/armtrafficmanager v1.3.0
 	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.4
-	github.com/aws/aws-sdk-go-v2 v1.42.0
+	github.com/aws/aws-sdk-go-v2 v1.42.1
 	github.com/aws/aws-sdk-go-v2/config v1.32.7
 	github.com/aws/aws-sdk-go-v2/service/account v1.24.0
 	github.com/aws/aws-sdk-go-v2/service/amplify v1.38.13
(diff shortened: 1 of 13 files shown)


ztgrace:
@ai-review

ztgrace:
@ai-review

PR #255, opened 2026-07-27
LAB-4616: in-repo architecture docs and agent skills

ztgrace:
Closes LAB-4616.

Migrates Aurelian's architecture docs and agent skills into the repo ahead of the `praetorian-cloud` and `praetorian-core` retirement, and puts a CI gate on their accuracy. Harness-agnostic throughout — no `CLAUDE.md`, no vendor-specific syntax.

## What lands

| File | Role |
| --- | --- |
| `ARCHITECTURE.md` | Aurelian invariants — what is true and must be true about this system |
| `DEVELOPMENT.md` | Go craft invariants plus DRY/YAGNI/comments; lifts into any Praetorian Go repo |
| `AGENTS.md` | Vendor-neutral router per [agents.md](https://agents.md/); points, never inlines |
| `.agents/skills/*/SKILL.md` | Three [Agent Skills](https://agentskills.io) workflows: module development, integration testing, code review |
| `scripts/check-architecture-refs.sh` + `docs-check.yml` | Freshness gate and skill-format validation |
| `CODEOWNERS` | Extended to the new rule paths |

## The organising rule

Anything true both while building **and** while reviewing is an invariant. It lives in `ARCHITECTURE.md` (Aurelian) or `DEVELOPMENT.md` (Go craft), in exactly one of them, and skills cite it rather than restate it. A CI audit asserts each invariant appears in exactly one file.

That is the fix for the actual problem: the rules previously existed in three places — a private skill library, a ~200-line inline prompt in `ai-review.yml`, and 16 lines of README — and had already drifted apart.

## Corrections to live behaviour

Every symbol was verified against `pkg/` rather than transcribed. Two of the errors found are **currently active in `ai-review.yml`**, meaning the reviewer has been generating findings against correct code:

- **`output.CloudResource` does not exist.** The prompt instructs "modules emit `output.CloudResource` only". `grep -rn '^type CloudResource' pkg/` returns nothing. Same for `SecretFinding` (the real type is `secrets.SecretScanResult`).
- **Double-close is safe, not a defect.** The prompt lists "double-close on the same pipeline" as a Tier 1 bug. `pkg/pipeline/pipeline.go:59` documents `Close` as *"safe to call multiple times"* and guards it with `closeOnce`. Send-after-close is real and is kept.

Also corrected from the source material: `ratelimit.NewPaginator` (constructors are per-CSP), the nebula-era `modules/aurelian/test/terraform/` fixture path, a nonexistent exported GCP ID parser, an `output.ScanInput`/`BaseAurelianModel` overclaim, an enricher-registry undercount, and a shared-container `TestMain` pattern asserted in bold by the source that has **zero occurrences** repo-wide.

`strings.Split` on resource identifiers is scoped to new and changed code — 18 existing violations mean an absolute rule would fire on every touched legacy file.

## capability-sdk

`ARCHITECTURE.md` §8 documents the migration to `capability-sdk/pkg/capmodel` as **direction, with no rule attached** — it has not reached this repo (`go.mod` has no requirement, the tree has no `capmodel` references), so a rule would flag correct code today. The enforceable part is §10 item 6: no `capability-sdk` import under `pkg/modules/`, preserving the migration plan's own constraint that Aurelian stays a standalone binary.

## Not in this PR

- **`ai-review.yml` is untouched.** It is rewired in a follow-up to read these files from `main` via a base-ref checkout, which requires them to be on `main` first.
- **CODEOWNERS is advisory until `require_code_owner_review` is enabled.** All three rulesets carrying a `pull_request` rule have it `false`, and two are Organization-scoped — so that is not a repo-side change.

## Verification

```
scripts/check-architecture-refs.sh   OK: all doc references resolve
skills validation                    validated 3 skill(s)
DRY audit                            every invariant in exactly one file
```

### DIFF
--- a/.agents/skills/aurelian-code-review/SKILL.md
+++ b/.agents/skills/aurelian-code-review/SKILL.md
@@ -26,6 +26,15 @@ finding — a paraphrase drifts from the source and the author cannot contest it
 - Flag only what you can pin to a concrete file and line.
 - Flag only what is observable in the diff.
 
+Adjacent pre-existing issues are the one exception, and they are **not** findings. If
+the diff leads you to a real problem in untouched code, put it in Notes, say it is
+pre-existing and out of scope, and name the file. It does not count against the caps
+and does not change the verdict. Two reasons: the observation is worth keeping as
+backlog rather than discarding, and recording it separately stops it leaking into the
+findings table where it would block a PR that did not cause it. If it genuinely blocks
+this change — the diff cannot be correct until it is fixed — then it is in scope, and
+it goes in the table as a finding like any other.
+
 ## Procedure
 
 1. Read the PR title and body.
@@ -43,11 +52,21 @@ principle finding.
 Spend most of the budget here. These are review-only judgments: no document can state
 them in advance because they depend on what the diff does.
 
+**Every list in this pass is illustrative, not exhaustive.** The headings name the
+classes of defect worth hunting; the bullets are the shapes seen most often in this
+codebase. A defect that changes behaviour belongs in this pass whether or not it
+matches a bullet. Do not treat a list as a checklist you can complete, and never
+withhold a real finding because it is not enumerated.
+
 ### Intent
 
 Does the change do what the PR body and the changed function names claim? Flag scope
-creep beyond stated intent. A title-versus-code mismatch is not itself a defect — flag
-it only when the divergence introduces real risk.
+creep beyond stated intent.
+
+A PR body is a snapshot: authors push commits without rewriting it, so it is routinely
+incomplete rather than wrong. Silence in the body is not a finding. Flag only when the
+code **contradicts** a stated claim, or when work outside the stated scope carries real
+risk. Never open a finding whose remedy is "update the PR description".
 
 ### Logic
 
@@ -67,7 +86,8 @@ it only when the divergence introduces real risk.
 - A local variable shadowing an imported package name — `pipeline`, `context`,
   `errors`.
 
-The rules behind these are `DEVELOPMENT.md` §4. Your job is spotting the instance.
+The rules behind these are `DEVELOPMENT.md` §4, which you have already read. Cite the
+section; do not restate it. Your job is spotting the instance in the diff.
 
 ### Panic risk
 
@@ -180,4 +200,8 @@ Caps:
 - Quality over quantity. If a finding would not survive a human reviewer's pushback,
   omit it.
 
-If nothing significant: `No issues found — LGTM pending human review.`
+Use `No issues found — LGTM pending human review.` only when pass 1 produced nothing
+at all and passes 2 and 3 produced nothing above low severity. A low-severity finding
+still goes in the table; it does not by itself withhold the line. Anything in pass 1,
+or any medium or high from passes 2 or 3, means findings — not LGTM. Notes entries do
+not affect this: pre-existing observations are compatible with LGTM.
--- a/.agents/skills/aurelian-code-review/SKILL.md
+++ b/.agents/skills/aurelian-code-review/SKILL.md
@@ -1,0 +1,183 @@
+---
+name: aurelian-code-review
+description: Use when reviewing an Aurelian pull request. Covers the behavioral defect pass, the architecture invariant pass, and the principles pass for Go cloud recon code across AWS, Azure, GCP, and Kubernetes.
+license: Apache-2.0
+---
+
+# Aurelian code review
+
+Advisory. Technical feedback only. You do not approve or reject — the human reviewer
+decides.
+
+Two normative documents hold the rules. This skill does not restate them:
+
+- `ARCHITECTURE.md` — Aurelian invariants, sections 1 through 11.
+- `DEVELOPMENT.md` — Go craft and the three enforceable principles, sections 1 through 9.
+
+Read both before reviewing. Cite them by section. Never paraphrase a rule into a
+finding — a paraphrase drifts from the source and the author cannot contest it.
+
+## Scope discipline
+
+- Code review only. Do not demand new tests. Do not comment on coverage, missing
+  tests, or test naming.
+- No stylistic, formatting, or naming nits beyond the low-severity list below.
+- Do not propose refactors outside the diff. Do not edit files. Do not push commits.
+- Flag only what you can pin to a concrete file and line.
+- Flag only what is observable in the diff.
+
+## Procedure
+
+1. Read the PR title and body.
+2. Read the diff.
+3. Read each changed file **in full**. Never review a diff in isolation — the
+   surrounding function decides whether a diff-local line is a defect.
+4. Run the three passes below, in order.
+5. Post one top-level comment. No inline comments.
+
+Order matters. A behavioral defect outranks an invariant deviation, which outranks a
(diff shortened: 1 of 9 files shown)


ztgrace:
@ai-review

Note for the reviewer: this PR is reviewed by the **current** inline prompt, which contains two rules that this PR removes as factually wrong. Expect possible false findings on those two points specifically:

1. The prompt says *"modules emit `output.CloudResource` only"*. That type does not exist in `pkg/output/`. `ARCHITECTURE.md` §8 documents the real types and states explicitly that `CloudResource` and `SecretFinding` do not exist.
2. The prompt lists *"double-close on the same pipeline"* as a Tier 1 defect. `pkg/pipeline/pipeline.go:59` documents `Close` as safe to call multiple times, guarded by `closeOnce`. The new review skill keeps send-after-close, which is a real panic, and adds an explicit do-not-flag note for double close.

The diff is markdown, bash, and YAML — no Go changes — so most Go-specific rules should not apply.

.agents/skills/aurelian-code-review/SKILL.md
```diff
@@ -0,0 +1,183 @@
...
+principle finding.
+
+## Pass 1 — behavioral defects
+
+Spend most of the budget here. These are review-only judgments: no document can state
+them in advance because they depend on what the diff does.
+
+### Intent
+
+Does the change do what the PR body and the changed function names claim? Flag scope
```
xoverride, line 63: This will largely flag issues. As we make changes and commit, it might not necessarily come with a PR body update. 
We can either consider rewording this pass. Or enforcing some hooks, that ensures updating a PR description when push if one exist (but this might be fragile and need considerations).

.agents/skills/aurelian-code-review/SKILL.md
```diff
@@ -0,0 +1,183 @@
...
+creep beyond stated intent. A title-versus-code mismatch is not itself a defect — flag
+it only when the divergence introduces real risk.
+
+### Logic
+
+- Off-by-one, wrong loop bounds, wrong slice indices.
```
xoverride, line 73: While those look reasonable, I'm a bit worried about giving specific logic cases here would limit the module's capability and coverage.

.agents/skills/aurelian-code-review/SKILL.md
```diff
@@ -0,0 +1,183 @@
...
+
+- A closure capturing the wrong variable when the closure is created outside a `for`
+  statement. The for-loop case is fixed by the language; closures built in other
+  constructs still carry the risk.
+- Two goroutines writing the same field with no synchronization.
+- Cancellation dropped on a long-running call.
+- A local variable shadowing an imported package name — `pipeline`, `context`,
+  `errors`.
+
+The rules behind these are `DEVELOPMENT.md` §4. Your job is spotting the instance.
```
xoverride: Again, if we are referencing the development.md document here and prior instruction have instructed to read the two documents. We should expect the module to already have context of those two documents. So we should either consider dropping the section completely or only mention the reference instead of giving a subset of examples.

.agents/skills/aurelian-code-review/SKILL.md
```diff
@@ -0,0 +1,183 @@
...
+| Parameter binding, tags, post-bind | §6 |
+| Enricher registration, mutator and evaluator split | §7 |
+| Output types | §8 |
+| Pagination, region fan-out, rate limiting, keyed state | §9 |
+| The six prohibitions | §10 |
+| Load-bearing patterns — a carve-out, never a finding | §11 |
+
+§4 is the highest-value check in this pass; it is the codebase's primary deadlock
+source. §10 prohibition 5 binds new and changed code only — the tree has pre-existing
+violations, so do not flag untouched lines.
```
xoverride, line 155: Just a thought. Often times during review, it might find existing issues in the codebase that are related to the PR. Sometimes that is blocking, which would require resolving before the PR can continue. And sometimes it might be non-blocking. While I agree that the PR review should be focused on the PR changes only for the purpose it is made for. It might be still valuable to capture what it has seen as a side artifact for the non-PR-related issues and open up issue as blacklog.

.agents/skills/aurelian-code-review/SKILL.md
```diff
@@ -0,0 +1,183 @@
...
+detail. Optional notes last.
+
+Caps:
+
+- Under 15 findings.
+- At least two-thirds from pass 1, unless the PR is purely a convention fix.
+- Quality over quantity. If a finding would not survive a human reviewer's pushback,
+  omit it.
+
+If nothing significant: `No issues found — LGTM pending human review.`
```
xoverride: I would be explicit on when it is allowed to state the LGTM line. 
What kind of issue will be considered significant and what will not?

.agents/skills/aurelian-integration-testing/SKILL.md
```diff
@@ -0,0 +1,360 @@
...
+## Conventions
+
+| Item | Pattern |
+| --- | --- |
+| Test file | `<module_name>_integration_test.go`, in the same package as the code under test |
+| Build tag | `//go:build integration` — first line, blank line, then `package` |
+| Fixture directory | Mirrors the module path. See ARCHITECTURE.md §2. |
+| Fixture path argument | Relative to `test/terraform`, e.g. `"aws/recon/graph"` |
+
+## Workflow
```
xoverride, line 31: While the skill description says the skill is about writing or modifying integration tests for Aurelian. The workflow here seems to also imply the execution part of the integration test. I am not sure whether that is intended or there will be a dedicated integration execution skill specifically for that.

.agents/skills/aurelian-module-development/SKILL.md
```diff
@@ -0,0 +1,168 @@
...
+   emit. Keep it thin.
+
+8. **Choose what `Run` returns using the table in ARCHITECTURE.md §4 ("What Run
+   returns").** This is the single most common source of deadlocks and truncated
+   output in this codebase. The right answer depends on who closes the output
+   pipeline, and there are three distinct cases. Read the table; do not guess.
+
+9. **Emit an output type from the §8 table.** Do not invent a result type. New
+   findings use `output.AurelianRisk`.
+
```
xoverride, line 71: I think quite a few of those rules instructed here can be statically enforced. I will take that as a different task.

ARCHITECTURE.md
```diff
@@ -0,0 +1,366 @@
...
+
+Dependency direction, stated as "imports" rather than arrows:
+
+- `cmd/` imports `pkg/modules/<csp>/<category>/` — blank imports, for `init()` registration —
+  plus `pkg/plugin`, `pkg/output`, `pkg/model`.
+- `pkg/modules/` imports `pkg/plugin`, `pkg/<csp>/<service>/`, `pkg/output`, `pkg/pipeline`.
+- `pkg/plugin` imports `pkg/output`, `pkg/model`, `pkg/pipeline`, `pkg/templates`, `pkg/graph`,
+  and some `pkg/<csp>/<service>/` packages. It **never** imports `pkg/modules/`.
+
+Two rules follow, and both hold today:
```
xoverride, line 43: I think those rules can also be statistically enforced.

xoverride (commented):
WIP

PR #260, opened 2026-08-28
feat(aws): add ECR dump recon module

ztgrace:
> **Merge #261 into this branch first.** It carries the completion work: a
> framework parameter-leak fix that a live integration run exposed, a path
> traversal fix in layer extraction, and every verified reviewer finding below.
> Once #261 lands here, this PR gets a full CI pass over the combined code.

Reopens #116, which GitHub refuses to reopen: the original branch predates the
history squash and shares no ancestry with today's `main`, and once rebased it
counts as force-pushed. Aarushi's three commits are replayed onto current `main`
byte-for-byte with authorship intact; the old tip is preserved at
`aarushi/ecr-dump-pre-rebase`.

Authored by @aarushidwi. The trailing `chore` commit is only the rebase
adaptation.

## Summary
- Ports Nebula's `ecr-dump` module to Aurelian with full feature parity
- Lists and pulls images from both private and public ECR repositories
- Extracts container layers and scans for embedded secrets using Titus
- Uses `go-containerregistry` instead of Docker daemon (no daemon dependency)
- Console output grouped by rule name, matching Nebula's `NPFindingsConsoleOutputter` format

## Nebula Parity

| Feature | Nebula | Aurelian |
|---------|--------|----------|
| Private ECR repos | ✅ | ✅ |
| Public ECR repos | ✅ | ✅ |
| Container image pull | Docker SDK | go-containerregistry (no daemon) |
| Layer extraction | Docker SDK | go-containerregistry |
| Secret scanning | NoseyParker | Titus (`secrets.SecretScanner`) |
| Console output (grouped by rule) | ✅ | ✅ |
| JSON output (`AurelianRisk`) | N/A | ✅ |
| `extract` parameter (layer files) | ✅ | ✅ |
| Multi-region support | ✅ | ✅ |

## Rebase adaptation (`chore` commit)

`main` moved underneath this branch; the module needed these to work against it:

- **`secrets.NewSecretRisk` no longer exists.** Risks now go through
  `SecretScanResult.ToRisk()`, and the module's private `buildECRProofData` is
  deleted — it duplicated `pkg/secrets/risk.go` and dropped the match location
  and validation fields the proof format now carries.
- **`ScanInput.Platform` is now required for risk naming.** Without it every
  risk was named `-secret-*` instead of `aws-secret-*`.
- **Drain `SecretScanner.Flush()` after the last `Scan`.** Matches Titus
  withholds behind its regexp-timeout retry queue were being silently dropped;
  every other secret-scanning module gets this via `ScanFlushAndClose`.
- **Classified `ECR` as reviewed-regional** in `pkg/aws/resourcetypes`. This
  module's `SupportedResourceTypes()` introduces the service to the `GetAll()`
  union, and main's drift guard fails any unclassified service.
- Added a unit test for the risk conversion, dropped find-secrets' now-stale
  `// TODO: AWS::ECR::Repository — deferred to follow-up PR`, regenerated CLI docs.

## Image selection

`--images` takes `latest` (default), `newest`, `all`, or any other tag name.

`latest` is the default because that is the tag every container client resolves
when a reference carries none, so it is what an operator means by "the" image in
a repository. It is **not** a synonym for the most recently pushed image: a
release tagged `v3` pushed after `latest` leaves the two pointing at different
images. `newest` selects by push time for the cases where that is what you want.

A repository with **no** `latest` tag falls back to its most recently pushed
image, logged. Version-tagged registries commonly have no `latest` at all, and
silently skipping them would drop most of such a registry from a secret scan.

The default costs no extra API call: the enumerator already lists every image to
compute `LastModified`, so it publishes the `latest`-tagged reference alongside
the newest one and the module reads whichever it needs. `LastModified` keeps
tracking the newest push, since that is what bounds the checkpoint.

Selection composes with the checkpoint at **image** granularity: an incremental
`--images all` re-scans only images pushed after it, while the repository-level
check still short-circuits first, so an unchanged repository is skipped without
listing its images at all.

Two collisions that only `all` exposes are fixed:

- **Extraction is now per image** (`<repo>/<tag>/layer0/…`). Layers are numbered
  per image, so two images of one repository both wrote `layer0/` into the same
  directory and clobbered each other.
- **`ScanInput.Label` is prefixed with `repo:tag`** (or `repo@digest` for an
  untagged image). Without it, findings from two images of one repository were
  indistinguishable.

Also: the private and public registries return different SDK types, so both now
map onto `imageCandidate` and share one `chooseImages` policy — which means the
public-ECR path honours `--images` too, and the duplicated newest-image
resolution is gone. Untagged images are referenced by digest instead of dropped.

Reserved-word caveat: a repository with an image literally tagged `newest` or
`all` cannot select it by tag.

## Incremental scanning (parity with #258)

find-secrets got `--modified-since` in #258 via the shared AWS extractor's
`LastModified` check. ecr-dump enumerates ECR itself and never passes through
that extractor, so the same contract is implemented here:

- `--modified-since <RFC3339>` skips any repository whose newest image was
  pushed no later than the checkpoint, before the pull. An unchanged registry
  costs one `DescribeImages` per repo instead of a full layer download and scan.
  `ImagePushedAt` is the authoritative AWS modification time, the analogue of
  `AWSResource.LastModified`.
- An image with **no** push timestamp is never skipped — missing modification
  metadata fails open, matching the extractor's nil-`LastModified` behaviour.
- `SetFailOnError(incremental)` plus fail-closed handling of every path that
  previously warned and carried on (region config, DescribeRepositories, auth,
  DescribeImages, layer read/extract, scan, Titus flush and close). A caller
  that advances its checkpoint must never be told that a run which silently
  skipped content succeeded. Best-effort runs keep warn-and-continue.
- An empty repository is now distinguished from a `DescribeImages` failure
  (`errNoImages`); previously both logged "no images" and were ignored, hiding
  access-denied errors.

## Listing goes through the shared enumerator

The skip above only works where a **native** enumerator supplies the timestamp —
CloudControl's `AWS::ECR::Repository` description carries no modification time,
which is what #258's *"route secret scans through native enumerators"* commit was
about. ecr-dump was also hand-rolling its own `DescribeRepositories` loop, so:

- New `ECRRepositoryEnumerator` in `pkg/aws/enumeration`, modelled on the four
  timestamp-bearing enumerators #258 added. It resolves each repository's newest
  image, publishes `ImagePushedAt` as `LastModified`, and publishes the pull
  reference and tag in `Properties` so consumers don't repeat `DescribeImages`.
  Registered on the dispatcher and added to
  `TestNewEnumerator_RegistersTimestampEnumerators`.
- A `DescribeImages` denial is recorded as a `SkipReport` entry and the repo is
  still emitted with `LastModified` nil, so a modified-since consumer fails open
  and scans it instead of treating it as unchanged.
- ecr-dump lists through the shared `Enumerator`, which also gets it ARN
  targeting (`--resource-arn`) and the same incomplete-enumeration gate
  find-secrets has: an incremental run fails if any repository was skipped.
- Deletes `listECRRepos` and `getLatestImage` from the module, drops the
  `ProxyEndpoint` account-ID parsing (`AccountRef` arrives on the resource), and
  uses the enumerated repository ARN rather than rebuilding it. The registry
  authenticator is now cached per region — `GetAuthorizationToken` issues a
  12-hour token, so one call per region covers a whole run.
- `AWS::ECR::PublicRepository` is excluded from the list-all union: ECR Public
  has a single us-east-1 control plane, so a per-region CloudControl fan-out
  would query it in ~30 regions where the API doesn't answer. The module still
  enumerates public repositories itself, since that registry doesn't fit the
  per-region enumerator model.

Net effect on the module: **-17 lines** despite gaining the feature.

## New Dependencies
- `github.com/aws/aws-sdk-go-v2/service/ecr`
- `github.com/aws/aws-sdk-go-v2/service/ecrpublic`
- `github.com/google/go-containerregistry` (promoted from indirect)

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go vet -tags integration ./...` clean
- [x] `go test ./...` passing
- [x] `golangci-lint run` clean on `pkg/modules/aws/recon` and `pkg/aws/resourcetypes`
- [x] Unit tests: registration, metadata, parameters, `sanitizeName`, `isBinary`, `emitResults` risk shape, `skipUnchanged` truth table (including fail-open on nil push time), fail-closed `problem`, invalid `--modified-since` rejection, `processRepository` skip paths that must not reach AWS, `chooseImages` selection matrix (9 cases incl. per-image checkpoint filtering and the no-`latest` fallback), `selectImages` published-reference fast paths (asserted to reach no AWS client), `imageReference` tag/digest fallbacks, `extractLayer` label scoping and on-disk layout; enumerator tests for `buildECRRepositoryResource` (tagged, untagged, no-images, nil-ARN) and ARN validation
- [x] Live tested against firing range (account 555045483459) — 5,218 findings across 15 rules from 2 repos (pre-rebase)
- [x] **Integration suite green against real AWS** — default profile, account 411435703965, us-east-2; all 11 `TestECRDumpIntegration` subtests plus `TestECRDumpEmptyRegistry`, provisioning and tearing down real ECR repositories and pushed images, no orphans left. Validated on the combined code with #261 applied. Reviewer findings are addressed in #261; see that PR for each one
- [x] Superseded note: previously unrun

## Follow-up worth considering (not in this PR)
`ScanInput.PathFilterable` is left `false` for layer files, so the ignore-pattern
machinery never fires on container filesystems — likely a large share of the
5,218-finding count is vendored dependencies. Setting it would need the `Label`
reworked from `repo:layerN/path` into a clean path, which changes the console
output format that currently mirrors Nebula.


ztgrace:
@ai-review

Rebased onto `main` (now includes the Front Door / S3 dnstakeover work from #265) and fixed the `ci / Lint` failure.

The lint job was failing before it ran a single check:

```
can't load config: the Go language version (go1.26) used to build golangci-lint
is lower than the targeted Go version (1.27.0)
golangci-lint exit with code 3
```

Cause was a titus bump to `v1.2.10-0.20260911154726-bd2a668c13ac`, which declares `go 1.27.0` and dragged this module's own go directive up with it. CI's golangci-lint 2.12.2 is built with go1.26 and refuses to load the module. Nothing in the ECR work uses the newer titus, so it's pinned back to `v1.2.6` (matching main) and the directive is back to `go 1.25.8`; `go mod tidy` drops two indirect deps that only the newer titus pulled in.

Verified locally on the rebased head:
- `go build ./...` — OK
- `golangci-lint run --path-mode=abs --timeout=10m` — 0 issues
- `go test -race -coverprofile=coverage.out -covermode=atomic ./...` — 54 packages pass, 0 failures
- `go mod tidy` stable, `go mod verify` OK

ztgrace:
@ai-review

